PR #2, opened 2015-01-21
Update Vagrantfile

humaton:
group docker is automaticaly created with installation of docker package.


langdon:
Bug must have been fixed. This was a work around :)

Issue #5, opened 2015-03-02
Proposal: Document workflow for a developer for developing on OCP using CDK/ADB


ncoghlan:
Why? Because this subproject should be about more than shipping a bunch of components as a bundle - it should be about tying them together into a coherent workflow for the development of containerised services intended for eventual deployment on Atomic Host (and potentially other environments).

goern:
+1

langdon:
@ncoghlan  Let me fix your #3 and then see if you still think "workflow" is the right word :)

ncoghlan:
I suspect that's going to depend significantly on how much scope creep you decide to let me get away with, as I have some expansive ideas :)

ncoghlan:
Hmm, after seeing the updated README, perhaps my "suggested workflow" idea
should be a _separate_, documentation only project. So the bundle delivers
a box of parts, and then a separate project called something like "Atomic
Suggested Workflow" would describe one possible way of stringing them
together into a cohesive developer experience.

Does that approach sound plausible to you?

ncoghlan:
As indicated by the title change, I realised this idea can likely be tackled within the context of the Developer Bundle by providing an _example_ workflow that shows how to integrate an Atomic Developer Bundle based local development workflow with various CI systems and deployment environments to create an end-to-end code-creation -> production-deployment pipeline.

So it wouldn't even need to rise to the level of being a "Suggested" workflow - it could just be an example of how the bundle can be combined with a specific selection of other infrastructure management components (e.g. invoking Ansible from the Dockerfile to script container builds or from the Vagrantfile to script VM builds, accessing your local source repos we're you're modifying code, etc). Other folks would swap in whatever they're using (e.g. Salt, Chef, or Puppet instead of Ansible, drone.io instead of Jenkins for CI, etc) in place of what's used in the example workflow.

ncoghlan:
Revised the title again, as there's no need to restrict this to just one example workflow - we could have multiple examples, demonstrating integration with different systems. So you could have, for example:
- a "Gerrit review, Jenkins CI, Ansible CM, oVirt deployment" workflow
- a "Kallithea review, Buildbot CI, Salt CM, OpenStack deployment" workflow
- a "GitHub review, Travis CI, virtualenv CM, OpenShift Online deployment" workflow

and so-on and so forth. You wouldn't want to demonstrate every possible combination, but even just the three above examples would cover a lot of ground. Does that make sense?

[...]

PR #6, opened 2015-03-02
Add a project rationale to the README

ncoghlan:
After reading http://fishbowl.pastiche.org/2015/03/02/vagranthowto/, I thought it summed up the rationale for this project so well, we needed to reference it _somewhere_.

Note that the heading in this PR assumes the alternative name suggestion in #5 is accepted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-cdk
-===
+Atomic Developer Workflow
+=========================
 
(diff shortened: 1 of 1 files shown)


langdon:
I think the README has significantly changed, hopefully for the better, so closing this PR.

PR #11, opened 2015-08-04
Moving adb v2 initial code to master

LalatenduMohanty:


langdon:
LGTM for the initial adb-v2

PR #48, opened 2015-08-05
add virtualbox support to vagrantfile -- issue 46

jasonbrooks:
This vagrantfile does work w/ libvirt and virtualbox: added bit to set RAM at 1024 in vbox, changed name to just "Vagrantfile", vagrant up --provider=libvirt and vagrant up --provider=virtualbox both work.


LalatenduMohanty:
LGTM

PR #56, opened 2015-08-06
Changing the default hostname to centos7

LalatenduMohanty:

### DIFF
--- a/build_tools/kickstarts/centos-7-adb-vagrant.ks
+++ b/build_tools/kickstarts/centos-7-adb-vagrant.ks
@@ -4,7 +4,7 @@ text
 keyboard us
 lang en_US.UTF-8
 skipx
-network --device eth0 --bootproto dhcp
+network --device eth0 --bootproto dhcp --hostname centos7-adb
 rootpw vagrant
 firewall --disabled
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
I need to test the patch

langdon:
You don't want to just do this in the Vagrantfile? LIke does having a hostname in the kickstart matter?

LalatenduMohanty:
@whitel I thought if we can change the hostname in the ks file then people using atlas.hasicorp also will also get benefited. Thoughts?  /.CC @kbsingh

langdon:
I don't think it matters much either way, just pointing out you can do it in the vagrantfile.

bexelbie:
The Centos box on atlas doesn't reset the hostname.  So I am not sure this is required.  Additionally, if we are going to set a hostname I would encourage <OS>-ADB so it is clear it is not a base box.  In this case CentOS7-ADB

bexelbie:
Per conversation with @LalatenduMohanty and Navid on 13 October - patch will be rebased and updated with "CentOS7-ADB" and merged.

bexelbie:
LGTM

PR #59, opened 2015-08-07
[WIP] Steps to setup the development environment for packaging using atomicapp

LalatenduMohanty:

### DIFF
--- a/docs/setup-development-environment.rst
+++ b/docs/setup-development-environment.rst
@@ -1,0 +1,30 @@
+=========================================
+Development environment for packaging apps using Atomicapp
+=========================================
+
+<work in progress>
+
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
Do we need something like this as doc?

langdon:
So, yes, but it seems a little out of left field. I think we need a more intro doc too. Something like "we recommend packaging container applications up using nulecule. See the spec [here](example.com)." and then go in to the how to. Maybe we can ask docs to take a look at the content here and think about it in general. @bexelbie are you still the right person to ask?

bexelbie:
I am now.  I believe Atomic App is being rewritten with an API that may change the way the user uses the box.  I am starting a mailing list thread, but would appreciate your input in filling out how the user uses the ADB.  Please see: http://etherpad.osuosl.org/ADB-Story

LalatenduMohanty:
Closing this PR as I dont think it is necessary

Issue #75, opened 2015-08-21
Pods are not created with present setup of kubernetes-1.0.0 in atomicapp/dev


navidshaikh:
With latest atomicapp/dev box, I am unable to create the pods

``` bash
[vagrant@localhost ~]$ rpm -q docker kubernetes
docker-1.7.1-108.el7.centos.x86_64
kubernetes-1.0.0-0.8.gitb2dafda.el7.x86_64

[vagrant@localhost ~]$ kubectl get nodes
NAME        LABELS                             STATUS
127.0.0.1   kubernetes.io/hostname=127.0.0.1   Ready

[vagrant@localhost ~]$ kubectl get rc
CONTROLLER   CONTAINER(S)   IMAGE(S)   SELECTOR   REPLICAS

[vagrant@localhost ~]$ kubectl get pods
NAME      READY     STATUS    RESTARTS   AGE

[vagrant@localhost ~]$ kubectl run nginx --image=nginx
CONTROLLER   CONTAINER(S)   IMAGE(S)   SELECTOR    REPLICAS
nginx        nginx          nginx      run=nginx   1

[vagrant@localhost ~]$ kubectl get pods
NAME      READY     STATUS    RESTARTS   AGE

[vagrant@localhost ~]$ # waited for few mins, no activity

[vagrant@localhost ~]$ kubectl get pods
NAME      READY     STATUS    RESTARTS   AGE

[vagrant@localhost ~]$ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES

[vagrant@localhost ~]$ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             VIRTUAL SIZE

[vagrant@localhost ~]$ systemctl status kube-apiserver
kube-apiserver.service - Kubernetes API Server
   Loaded: loaded (/etc/systemd/system/kube-apiserver.service; enabled)
   Active: active (running) since Fri 2015-08-21 05:05:34 EDT; 39min ago
     Docs: https://github.com/GoogleCloudPlatform/kubernetes
 Main PID: 16063 (kube-apiserver)
   CGroup: /system.slice/kube-apiserver.service
           └─16063 /usr/bin/kube-apiserver --logtostderr=true --v=0 --etcd_servers=http://127.0.0.1:2379 --address=127.0.0.1 --allow_privileged=false --service-cluster-ip-range=10.254.0.0/16 --admission_contr...
```

[...]

PR #76, opened 2015-08-21
Updates kube-apiserver configuration for KUBE_ADMISSION_CONTROL

navidshaikh:
Fixes: https://github.com/projectatomic/adb-atomic-developer-bundle/issues/75

 Removes `ServiceAccount` from the value of `KUBE_ADMISSION_CONTROL` variable
 in /etc/kubernetes/apiserver

### DIFF
--- a/build_tools/kickstarts/centos-7-kubernetes-vagrant.ks
+++ b/build_tools/kickstarts/centos-7-kubernetes-vagrant.ks
@@ -93,6 +93,14 @@ tuned-adm profile virtual-guest
 echo "%vagrant ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers.d/vagrant
 sed -i "s/^.*requiretty/#Defaults requiretty/" /etc/sudoers
 
+# Fixes issue #71
+mkdir -p /etc/pki/kube-apiserver/
+openssl genrsa -out /etc/pki/kube-apiserver/serviceaccount.key 2048
+
+sed -i.back '/KUBE_API_ARGS=*/c\KUBE_API_ARGS="/etc/pki/kube-apiserver/serviceaccount.key"' /etc/kubernetes/apiserver
+
+sed -i.back '/KUBE_CONTROLLER_MANAGER_ARGS=*/c\KUBE_CONTROLLER_MANAGER_ARGS="--service_account_private_key_file=/etc/pki/kube-apiserver/serviceaccount.key"' /etc/kubernetes/controller-manager
+
 #enable Kubernetes master services
 #etcd kube-apiserver kube-controller-manager kube-scheduler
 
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
@navidshaikh I dont understand why we need to remove "ServiceAccount". We need more details.

liggitt:
Removing it is not correct. The correct change is to set up the signing key for service account token generation

liggitt:
See https://github.com/kubernetes/kubernetes/issues/11355#issuecomment-127378691

navidshaikh:
@liggitt : ACK! 

Chose location for key at `/etc/pki/kube-apiserver/` rather using `/tmp/`, thoughts?

liggitt:
sure, whatever path is appropriate

LalatenduMohanty:
@navidshaikh do u have a scratch build with this patch? Also the patch looks like 2 commits. I think it should be just one commit as the previous commit not relevant anymore.

dustymabe:
@navidshaikh This looks wrong:

```
sed -i.back '/KUBE_API_ARGS=*/c\KUBE_API_ARGS="/etc/pki/kube-apiserver/serviceaccount.key"' /etc/kubernetes/apiserver
```

Shouldn't it be `KUBE_API_ARGS=--service_account_key_file=/etc/pki/kube-apiserver/serviceaccount.key` ?

jasonbrooks:
FWIW, this is how it looks when configured w/ https://github.com/kubernetes/contrib/tree/master/ansible:

```
KUBE_API_ARGS="--tls-cert-file=/etc/kubernetes/certs/server.crt --tls-private-key-file=/etc/kubernetes/certs/server.key --client-ca-file=/etc/kubernetes/certs/ca.crt --token-auth-file=/etc/kubernetes/tokens/known_tokens.csv --service-account-key-file=/etc/kubernetes/certs/server.crt"
```

LalatenduMohanty:
LGTM.  Scratch build by @navidshaikh at http://cbs.centos.org/koji/taskinfo?taskID=29206

LalatenduMohanty:
Scratch build is working fine.

```
 1  ls
    2  kubectl get nodes
    3  systemctl status docker
    4  atomic run projectatomic/helloapache
    5  kubectl get pods
    6  history 
    7  kubectl get pods
    8  curl localhost
```

```
[vagrant@localhost ~]$ kubectl get pods
NAME          READY     STATUS    RESTARTS   AGE
helloapache   1/1       Running   0          2m
```

vincentvdk:
Hi, 
Just stumbled on this thread because i have a similar issue deploying the load-balancer (kubernetes contrib). On the latest F23 images the ServiceAccount is left out. When adding it, i need to generate the certs and add 

```
KUBE_API_ARGS="--service_account_key_file=/path/secret.key"
```

I'm building my own Ansible playbooks so not a big deal, but I couldn't find this in the docs on the ProjectAtomic website. Also not sure if this should be the default or not.

PR #84, opened 2015-09-03
Refactored and updated docs

### LINKED ISSUE #79: Document manual update of the vagrant box from cloud.centos.org
Goes here: docs/updating-thebox.rst

bexelbie:
Converted all non-legacy docs to rst

Updated flow and language.  Better sectioning.

Fixes #79


LalatenduMohanty:
@bexelbie LGTM. Thanks for the change. 

However we might want to highlight few things . But we can discuss that and then do that next.

Issue #99, opened 2015-10-09
Vagrant image intermittently fails to bring up networking


LalatenduMohanty:
ADB box will have the same issue [1]  as CentOS base box as both have similar kickstart file. 
As @karmix  mentioned 

The vagrant box  intermittently fails to bring up networking when used in a configuration with more than one network adapter. This is because the 'eth0' connection profile is not associated with any specific interface. When NetworkManager tries to bring the 'eth0' connection up on a non-eth0 device (that is not on a network with a DHCP server), it logs a failure message such as:

network: Bringing up interface eth0: Error: Connection activation failed: Connection 'eth0' is already active on eth1

[1] reference : https://github.com/CentOS/sig-cloud-instance-build/issues/24

LalatenduMohanty:
The solution is to associate the eth0 in the Vagrant box to eth0 of the host machine

```
# Bind the eth0 connection to the eth0 device (#24)
nmcli connection modify eth0 connection.interface-name eth0
```

Thanks @karmix

praveenkumar:
@LalatenduMohanty Can we close this bug now since we already fixed it?

praveenkumar:
@LalatenduMohanty ping, what it your opinion on this issue?

LalatenduMohanty:
@praveenkumar I have not seen this occurring on ADB . So lets close this.

PR #100, opened 2015-10-09
Fixing networking issue with multiple nics in host

LalatenduMohanty:
Guthub issue #99


navidshaikh:
This patch needs to be tested on the machine with multiple nics before we merge it and also test if it does not break on single nic machine. A koji build with this patch should help.

bexelbie:
@lukaszachy can you test this on a machine with single and multiple NICs if @LalatenduMohanty gets you a build?

LalatenduMohanty:
Need to figure out if this effects Virtualbox as well as libvirt providers.

LalatenduMohanty:
Build of Vagrant box with the patch: https://cbs.centos.org/koji/taskinfo?taskID=50195

LalatenduMohanty:
Looks like we are fine without this patch . Hence closing this

PR #104, opened 2015-10-14
Updates for Fedora and CentOS Installation

bexelbie:


docs/installing.rst
```diff
@@ -2,8 +2,71 @@
...
+* Microsoft Windows and Mac OS X
+
+  The suggested virtualization provider is `VirtualBox <https://www.virtualbox.org/>`_.  Installation instructions are available `online <https://www.virtualbox.org/manual/UserManual.html>`_.  While the latest stable shipping release should work, the majority of testing has been done with version 5.0.0 on Mac OS X and **XXX** on Microsoft Windows.
+
+* Fedora
+
+  Two different virtualization providers are supported on Linux, `VirtualBox <https://www.virtualbox.org/>`_ and `libvirt <http://libvirt.org/>`_.  The choice as to which to use should be driven by your preferences and environmental concerns and is outside of the scope of this document.  Both will work equally well in their default configuration.  You may wish to read the section on file synchronization when making this decision.
+
+  * VirtualBox Installation instructions are available `online at the VirtualBox website <https://www.virtualbox.org/manual/UserManual.html>`_.
+
```
LalatenduMohanty, line 20: I think https://www.virtualbox.org/manual/ch02.html#startingvboxonlinux will be a better web link here.

docs/installing.rst
```diff
@@ -2,8 +2,71 @@
...
+    A summary of the installation is listed below:
+
+    ::
+
+      $ dnf install dkms
+      $ curl -O http://download.virtualbox.org/virtualbox/rpm/fedora/virtualbox.repo
+      $ mv virtualbox.repo /etc/yum.repos.d/
+      $ dnf install VirtualBox-4.3
+      $ sudo /etc/init.d/vboxdrv setup
+
```
LalatenduMohanty, line 30: Now that we have steps here, some would have the question with respect to why we need dkms and vboxdrv setup. So you may want to point them to https://www.virtualbox.org/manual/ch02.html#startingvboxonlinux :)

docs/installing.rst
```diff
@@ -2,8 +2,71 @@
...
+    ::
+
+      $ dnf install dkms
+      $ curl -O http://download.virtualbox.org/virtualbox/rpm/fedora/virtualbox.repo
+      $ mv virtualbox.repo /etc/yum.repos.d/
+      $ dnf install VirtualBox-4.3
+      $ sudo /etc/init.d/vboxdrv setup
+    
+    While the latest stable shipping release should work, the majority of testing has been done with version 4.3.30.
+
```
LalatenduMohanty, line 32: I would suggest that the tested version information should live in a separate page as t would be easy to maintain that. There is nothing wrong in putting it here, but it is easy to forget where we have the Virtualbox version mentioned in the docs (specially when have more documentation).
However its just a suggestion , If you think it should be here, I don't have any issue with it.

LalatenduMohanty:
@bexelbie we are adding more information about how to install Virtualbox and libvirt etc in the  "installing ADB" page.  May be we should move installation of virtualization technologies to a new page and then link it from installing ADB page. Because IMO the bigger page grows it kind of reduces the  readability. 
Thoughts?

docs/installing.rst
bexelbie, line 20: agreed.

docs/installing.rst
bexelbie, line 30: I'll change the link, but I am not going to expand the definition of VirtualBox to explain the need for dkms.  If you feel strongly about this, please open an issue and we can tackle it post-GA.

docs/installing.rst
bexelbie, line 32: For today, I'd like to leave it here.  Mostly because the current Vagrant for CentOS doesn't support the latest VirtualBox.  Once things are flowing more smoothly, we can fix that.

If I wind up refactoring this document into target based installation pages, I'll include the note only where needed and put the tested versions somewhere else.

bexelbie:
For now, I am leaving the pages long as I want to get the information into the repo first and then refactor later.

LalatenduMohanty:
@bexelbie , ok , I merging it. we should do he re-factor later (post GA)

PR #107, opened 2015-10-22
Fixing the doc issue in readme and installation guide (#106)

### LINKED ISSUE #106: [Doc] Instead of 'libvirt' it should be 'vagrant-libvirt' in readme
Currently README contain below step to start adb vagrant box if we use libvirt provider

```
Fedora 22: dnf install libvirt 
Which should be *dnf install vagrant-libvirt*
$ vagrant up
Which should be *$ sudo vagrant up*
```

Because if we only have libvirt installed and try to do **vagrant up** then it will fail with below error

```
$ vagrant up --provide libvirt
The provider 'libvirt' could not be found, but was requested to
back the machine 'default'. Please use a provider that exists.
$ rpm -qa libvirt
libvirt-1.2.13.1-3.fc22.x86_64
$systemctl status libvirtd.service 
● libvirtd.service - Virtualization daemon
   Loaded: loaded (/usr/lib/systemd/system/libvirtd.service; enabled; vendor preset: enabled)
   Active: active (running) since Wed 2015-10-21 06:57:57 EDT; 1h 9min ago
```

Please comment if I am missing something otherwise will correct it and send patch for same.

praveenkumar:
I updated doc a bit for better user understanding and this will fix #106.

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -30,7 +30,7 @@ Two virtualization providers have been tested with the ADB.
     
     While the latest stable shipping release should work, the majority of testing has been done with version 4.3.30.
 
-  * libvirt-provider is shipped in Fedora. Summary of installation is listed below:
+  * libvirt provider for vagrant is enabled using *vagrant-libvirt* package which is shipped in Fedora. Summary of installation is listed below:
     
     ::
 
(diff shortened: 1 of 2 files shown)


docs/installing.rst
```diff
@@ -56,14 +56,43 @@ Two virtualization providers have been tested with the ADB.
...
+        gpgcheck=0
+
+        [ror40-copr]
+        name=ror40-copr
+        baseurl=http://copr-be.cloud.fedoraproject.org/results/rhscl/ror40-el7/epel-7-x86_64/
+        enabled=1
+        gpgcheck=0
+
+      EOM
+
```
bexelbie, line 87: These changes are odd because you are suggesting that a vagrant Fedora COPR be used to install software that seems to have been available outside of that COPR.  Additionally you reference libvirt-provider (meaning vagrant-libvirt) but don't explain that.  A user would not initially understand what you are accomplishing as you seem to say one thing and do another.  If you really want to combine the vagrant and libvirt steps, you should combine the steps fully, not partially, imho.

LalatenduMohanty, line 87: Agree with  @bexelbie . @praveenkumar  all the changes looks fine except this one.

praveenkumar, line 87: @LalatenduMohanty Can you please elaborate more? Yesterday I had chat with @bexelbie and already made required changes. If their is any missing point I would love to add those information.

strzibny:
I am sorry I already committed the change `COPR -> SCL` so now this PR has conflicts.

bexelbie:
Fixed in #122

PR #116, opened 2015-11-02
Let vagrant-libvirt to pull in necessary dependencies

strzibny:
If `vagrant-libvirt` don't pull in all dependencies, it's a bug.

This simplifies the installation instructions and makes it clear that installation and running libvirt provider is actually easier.


bexelbie:
LGTM +1

PR #120, opened 2015-11-05
Removing the tls-certs-generation.sh as it is not used in the ks

LalatenduMohanty:
As the contents of the script are already used in the kickstart file


bexelbie:
LGTM

PR #121, opened 2015-11-10
Removing the single k8s setup as it is blocking OpenShift to run

LalatenduMohanty:


LalatenduMohanty:
scratch build works fine for me https://cbs.centos.org/koji/taskinfo?taskID=52561

LalatenduMohanty:
The single node setup  will be added as a script, so if someone needs a single node k8s setup , he/she should be able to get it. 

[1] https://github.com/LalatenduMohanty/k8s-setup/blob/master/k8s-setup.sh

bexelbie:
LGTM

navidshaikh:
Can we export the the k8s setup script inside the box (in KS file) and add required docs about it?

PR #122, opened 2015-11-10
Update docs to do vagrant-libvirt in a single cmd

### LINKED ISSUE #106: [Doc] Instead of 'libvirt' it should be 'vagrant-libvirt' in readme

bexelbie:
Fixes #106
Replaces #107

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,8 +44,7 @@ Detailed `installation instructions <docs/installing.rst>`_ are available.  Inst
 
    * For Linux, the recommended provider is `libvirt <http://libvirt.org>`_.  This is packaged for most distributions.
 
-     * Centos 7: ``yum install libvirt``
-     * Fedora 22: ``dnf install libvirt``
+     * `Install recommended provider for CentOS/Fedora <docs/installing.rst>`_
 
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
LGTM

PR #130, opened 2015-11-24
Removing the files as these are not required with current ADB

LalatenduMohanty:


LalatenduMohanty:
@whitel  I have removed the RHEL directory (the equivalent files are present in CentOS directory) . Also I have removed some Vagrantfiles as these are not required in ADB anymore because docker service is built-in to the box now. 
The reason of removing RHEL is also that , from upstream point of view, it would be easier to give CentOS support and maintain it . Let me know what you think

LalatenduMohanty:
@bexelbie @navidshaikh @strzibny comments?

navidshaikh:
LGTM

langdon:
@LalatenduMohanty  Where do you plan to provide the upstream version of the rhel vagrantfiles?

langdon:
@LalatenduMohanty also, is the plan to ignore the vagrant-docker provider option?

LalatenduMohanty:
@whitel I dont have any plan. Mostly because RHEL vagrant boxes are not availability in public.

With respect to vagrant-docker provider, we have relevant files in CentOS directory

LalatenduMohanty:
@whitel I have updated the patch . It is not much different than the previous one except I am not deleting the rhel directory. I have gone through the files and we are not deleting the docker provider stuff

langdon:
@LalatenduMohanty looks good combined with https://github.com/projectatomic/adb-atomic-developer-bundle/issues/133

LalatenduMohanty:
@whitel thanks. Merging this.

PR #168, opened 2015-12-18
Cleanup old ADB v1 files

bexelbie:


bexelbie:
While there is a story for testing apps on Atomic Host.  It seems these files are really just leftovers from ADBv1 and we should start fresh for this usercase.

@LalatenduMohanty @whitel ?

bexelbie:
@whitel PTAL

langdon:
They do slightly more than enable atomic-host. They also setup the docker-provider from vagrant for use on a reusable VM. Not that I think that is a complete reason for retaining them, I just want to clarify the difference. I am inclined to think this is not a use case we care about anymore, but, I thought it was worth mentioning.

bexelbie:
@whitel Perhaps they should be bundled in a repo/location for users of "blank" vms.  I don't think that is here.

LalatenduMohanty:
@bexelbie agree with you. LGTM

PR #171, opened 2015-12-18
[Do Not Merge] Documentation patch for Kubernetes Exposure

### LINKED ISSUE #169: Expose Kubernetes Service when Kube is running
The ports for Kube need to be exposed in the Kube single node setup.

The environment variable needed should be passed by vagrant-adbinfo - 

see: https://github.com/projectatomic/vagrant-adbinfo/issues/58

bexelbie:
This patch only fixes the docs - code needed
do not merge

Will Fix #169

### DIFF
--- a/components/centos/centos-k8s-singlenode-setup/README.rst
+++ b/components/centos/centos-k8s-singlenode-setup/README.rst
@@ -1,7 +1,11 @@
 Vagrantfile: CentOS Single Node Kubernetes
 ==========================================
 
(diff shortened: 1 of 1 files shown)


navidshaikh:
@bexelbie : I don't think we need any more code for this PR, what do you think is remaining ?

bexelbie:
@navidshaikh we need to determine if we have to expose KUBECONFIG

navidshaikh:
@bexelbie : ACK!

PR #187, opened 2016-01-11
Fix #183, added version info for ADB

### LINKED ISSUE #183: Add a version file in ADB/CDK
A version file will help identify the version of the box and help adbinfo to provide the information as suggested in https://github.com/projectatomic/vagrant-adbinfo/issues/64
 The file name can be something like /etc/adb-version or /etc/adb-release

praveenkumar:
We appended ADB version info to `os-release` so that `vagrant-adbinfo` consume it and provide required details to user.

### DIFF
--- a/build_tools/kickstarts/centos-7-adb-vagrant.ks
+++ b/build_tools/kickstarts/centos-7-adb-vagrant.ks
@@ -54,6 +54,10 @@ tuned
 sed -i "/HWADDR/d" /etc/sysconfig/network-scripts/ifcfg-eth*
 sed -i "/UUID/d" /etc/sysconfig/network-scripts/ifcfg-eth*
 
+# Add adb version info to consumed by adbinfo
+# https://github.com/projectatomic/adb-atomic-developer-bundle/issues/183
+echo "ADB_VERSION=\"2.0\"" >> /etc/os-release
+
 #Fixing https://github.com/projectatomic/adb-atomic-developer-bundle/issues/155
 echo "127.0.0.1     centos7-adb" >> /etc/hosts
 
--- a/build_tools/kickstarts/centos-7-adb-vagrant.ks
+++ b/build_tools/kickstarts/centos-7-adb-vagrant.ks
@@ -58,7 +58,7 @@ sed -i "/UUID/d" /etc/sysconfig/network-scripts/ifcfg-eth*
 # https://github.com/projectatomic/adb-atomic-developer-bundle/issues/183
 echo "VARIANT=\"Atomic Developer Bundle (ADB)\"" >> /etc/os-release
 echo "VARIANT_ID=\"adb\"" >> /etc/os-release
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
Wondering why /usr/lib/os-release file is not present in CentOS7 like Fedora

```
$ ls -l /etc/os-release 
lrwxrwxrwx. 1 root root 21 Oct 20 04:05 /etc/os-release -> ../usr/lib/os-release
```

LalatenduMohanty:
According to http://www.freedesktop.org/software/systemd/man/os-release.html  "/usr/lib/os-release" should exist and /etc/os-release  should be a symlink to it

bexelbie:
I believe we should also set the VARIANT information

praveenkumar:
> Wondering why /usr/lib/os-release file is not present in CentOS7 like Fedora

As per doc "The file `/etc/os-release` takes precedence over `/usr/lib/os-release`. Applications should check for the former, and exclusively use its data if it exists, and only fall back to `/usr/lib/os-release` if it is missing". So I think we are good as long as we wire up with /etc/os-release. It also point out `/usr/lib/os-release` is the recommended place to store OS release information as part of vendor trees.

> I believe we should also set the VARIANT information

What should be we set VARIANT here for CentOS (ADB)?

bexelbie:
My vote:

VARIANT="Atomic Developer Bundle (ADB)"
VARIANT_ID=adb

navidshaikh:
LGTM

build_tools/kickstarts/centos-7-adb-vagrant.ks
```diff
@@ -54,6 +54,12 @@ tuned
 sed -i "/HWADDR/d" /etc/sysconfig/network-scripts/ifcfg-eth*
 sed -i "/UUID/d" /etc/sysconfig/network-scripts/ifcfg-eth*
 
+# Add adb version info to consumed by adbinfo
+# https://github.com/projectatomic/adb-atomic-developer-bundle/issues/183
+echo "VARIANT=\"Atomic Developer Bundle (ADB)\"" >> /etc/os-release
+echo "VARIANT_ID=\"adb\"" >> /etc/os-release
+echo "VARIANT_VERSION=\"2.0\"" >> /etc/os-release
```
LalatenduMohanty: @praveenkumar I think the version should be 1.6.0. As Next release would be 1.6.0

praveenkumar: So we have different release version for CDK and ADB?

LalatenduMohanty: Yes, CDK and ADB have different versions at this point of time.

bexelbie:
LGTM, @praveenkumar have you built a scratch build?

praveenkumar:
@bexelbie Not yet, will discuss with @LalatenduMohanty  for scratch build.

LalatenduMohanty:
LGTM

Issue #140, opened 2015-12-09
Provide service scripts to control OpenShift on the VM


hferentschik:
Currently OpenShift can be using on the ADB VM using Vagrant provisioning. See #66 and the `Vagrantfile` provided via #ab4e2e3. 

There is no easy way for a user in this case to control (start/stop/reload) OpenShift apart of re-provisioning which is not intuitive. It also exposes a lot of low level details to the user.

A better approach would be to pull the required scripts into the VM itself and provide systmctl control scripts, so that OpenShift becomes a service like Docker. The creation of templates and users could still be done via Vagrant, since this might be user specific. The creation of a registry lies somewhere in between. Personally I think it could also be done as part of service scripts, since OpenShift is pretty much unusable without it.

Similar to the Docker service scripts, the OpenShift service script should define custom extension points for things a user most likely wants to change, eg the default route name.

In the Docker service configuration there are custom variables like `INSECURE_REGISTRY` and `ADD_REGISTRY` which are only used by the sysyemctl scripts and which modify the Docker daemon options if set.

hferentschik:
Please add label focus:devex

praveenkumar:
I started working towards it but still in very early stage and have to figure out some options before merge it to main repo.

https://github.com/praveenkumar/adb-atomic-developer-bundle/tree/origin_service/components/centos/centos-with-openshift/service-script

hferentschik:
> I started working towards it but still in very early stage 

Nice.

> and have to figure out some options before merge it to main repo.

I'll try to have a look and see how it works for our use case.

navidshaikh:
Having systemd service for OpenShift in the box should help identify if service is up and running in standard way, this in turn helps adbinfo plugin to check the service status and accordingly display connection information related: https://github.com/projectatomic/vagrant-adbinfo/issues/26

LalatenduMohanty:
Closing this as this issue is handled in https://github.com/projectatomic/adb-utils/issues

PR #191, opened 2016-01-12
Address #140 OpenShift service script on VM

praveenkumar:
This is address to use openshift as a service. Currently required origin/ose images should be available on target. To test it it contain `run.sh` which will place file in place and make sure you changes `openshift_option` according to your setup.


LalatenduMohanty:
@praveenkumar Looks good to me. However it looks like we cant merge the PR at this stage as it wont be included in the ADB ks file at this form automatically. Once we decide how we want to include it in ADB, then we need to modify the patch and then merge it.

praveenkumar:
Make sense, we will keep as it is till we have a plan around it.

LalatenduMohanty:
Closing this as this is handled by https://github.com/projectatomic/adb-utils

PR #196, opened 2016-01-15
Changed the state of packages installed via Ansible from `latest` to `present`

### LINKED ISSUE #193: Mesos provision script should not update package for subsequent run
Currently Mesos provision ansible playbook using `latest` which will make sure always updated packages and if some older version is available then it will update that. Assume a developer using a older version of mesos and deployed his application and meanwhile a new Mesos version is availabe which can break his application deployment then re-provision of vagrant box will be dangerous.

My vote is to changes `latest` to `present` in playbook and not auto-update any thing during subsequent provision.

```
          - name: Installing zookeeper, mesos, marathon, docker, git
            yum: name={{item}} state=latest
            with_items:
                - mesosphere-zookeeper 
                - mesos 
                - marathon
                - docker
                - git
```

dharmit:
Fixes #193 

@praveenkumar :smiling_imp:


LalatenduMohanty:
LGTM :+1:

praveenkumar:
@dharmit Thanks, LGTM :+1:

PR #223, opened 2016-02-05
Fix #222 httpd-tools package should be present in kickstart file

### LINKED ISSUE #222: httpd-tools package should be present in kickstart file
We require `httpd-tools` in our vagrant box because openshift provision require `htpasswd` utility which mentioned package provide to create authentication.

https://github.com/projectatomic/adb-utils/issues/3

praveenkumar:
This is required for openshift auth part if we are going to use `HTPasswdPasswordIdentityProvider` for identity.


LalatenduMohanty:
LGTM

PR #225, opened 2016-02-08
Set ip address for Marathon and Mesos-slave before installing packages.

kadel:
Packages are starting service, if we do config change after package is
installed we need that restart services. If we create configs before
installing pkgs, we don't have to restart them afterwards.

### DIFF
--- a/components/centos/centos-mesos-marathon-singlenode-setup/Vagrantfile
+++ b/components/centos/centos-mesos-marathon-singlenode-setup/Vagrantfile
@@ -21,10 +21,7 @@ Vagrant.configure("2") do |config|
     config.vm.provision "shell", inline: <<-SHELL
         sudo yum -y install epel-release
         sudo yum -y install ansible
-        sudo ansible-playbook /vagrant/provisioning/playbook.yml
-        sudo mkdir -p /etc/marathon/conf
-        sudo sh -c "echo \"#{IP_ADDRESS}\" > /etc/marathon/conf/hostname"
-        sudo sh -c "echo \"#{IP_ADDRESS}\" > /etc/mesos-slave/ip"
+        sudo ansible-playbook -e ip_address=#{IP_ADDRESS} /vagrant/provisioning/playbook.yml
     SHELL
 
     config.vm.network "private_network", ip: "#{IP_ADDRESS}"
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
@dharmit comments?

dharmit:
I somehow find it less than ideal to create configs before the install. Creating them post install and restarting service makes more sense. If that's not an issue than this looks good.

kadel:
We can do it with restart. But I don't like doing this, because for a moment you have running services that run with default (wrong) configuration and something could connect to it before you change config and restart it

dharmit:
That makes sense. What bothered me was that if in future, the package we're installing sets some default value in those files. Would the package installation override what we're doing using Ansible? But I tried for simple httpd package and that doesn't happen. A file created by user stays as is.

kadel:
I suppose that rpm shouldn't overwrite existing config files. But I don't know much about rpm packaging.

dharmit:
Nor do I. Perhaps @praveenkumar can answer that better. Out of the same concern, I said that creating a file before installation doesn't look like an ideal solutoin to me. :smile:

LalatenduMohanty:
@dharmit @kadel Whether RPM should overwrite a config file, that is dependent on the RPM i.e. RPM spec. However the in usual practice the config files does not replace the old files  but RPMs create <config>.new files or equivalent. There are some RPMs which replace existing config files and move the old config files to <config>.bak.

praveenkumar:
@dharmit, as @LalatenduMohanty  told, it's depend on rpm spec file. If you point out which package we are talking about I can tell you exactly if that will overwrite config file or not. As per best practice we should not overwrite config file of a package if we remove/uninstall but again it's depend upon how package is built. 

https://fedoraproject.org/wiki/Packaging:Guidelines#Configuration_files

dharmit:
Got it. Thanks for explaining.

@praveenkumar As far as the specific package is concerned, it's coming from the Mesosphere repo which can be enabled using http://repos.mesosphere.com/el/7/noarch/RPMS/mesosphere-el-repo-7-1.noarch.rpm

kadel:
It looks like that creating config files before installing packages should be OK.
I still think that this is better solution than restarting services during provisioning

LalatenduMohanty:
@kadel thanks. LGTM

LalatenduMohanty:
@navidshaikh @bexelbie @praveenkumar ping

bexelbie:
LGTM

PR #230, opened 2016-02-10
Fix #224 Provision fails after rebooting the machine

### LINKED ISSUE #224: Provision fails after rebooting the machine
From clean environment, if you do:

```
sudo vagrant up
sudo vagrant halt
sudo vagrant up
sudo vagrant provision
```

You'll get an error about already existing `openshift/origin` image.

I think line [69](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/components/centos/centos-openshift-setup/Vagrantfile#L69) doesn't do it's job as it effectively checks `:latest` image although the next command pulls `:v1.1.1` image.

praveenkumar:
Fix: #224 
This patch also carry updated openshift vagrant file.

### DIFF
--- a/components/centos/centos-openshift-setup/Vagrantfile
+++ b/components/centos/centos-openshift-setup/Vagrantfile
@@ -5,29 +5,18 @@
 DOCKER_REGISTRY="docker.io"
 
 # The name of the OpenShift image.
-ORIGIN_IMAGE_NAME="openshift/origin:v1.1.1"
+ORIGIN_IMAGE_NAME="openshift/origin"
+ORIGIN_TAG="v1.1.1"
 
 # The public IP address the VM created by Vagrant will get.
 # You will use this IP address to connect to OpenShift web console.
 PUBLIC_ADDRESS="10.1.2.2"
 PUBLIC_HOST="adb.cluster.io"
 
-# The directory where OpenShift will store the files.
-# This should be "/var/lib/openshift" in case you're not using the :latest tag.
-ORIGIN_DIR="/var/lib/origin"
-
 # The following steps are executed to provision OpenShift Origin:
 # 1. Create a private network and set the Vagrant Box IP to *10.1.2.2*. If you want a different IP then change the **PUBLIC_ADDRESS** variable.
 # 2. Pull latest the *openshift/origin* container from docker hub and tag it.
-# 3. Create the required configuration file directories which OpenShift Origin uses.  Also set the SELinux flag to make sure they will work with *Enforcing* mode.
-# 4. Run the OpenShift Origin docker container with various run options to make sure the required directory is mounted and the *host* network is used. **Note:** This takes about 15 seconds to start.
-# 5. Check if the OpenShift Origin container has started as expected.  If there is an error, output the docker logs.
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
LGTM. However how to make sure this does not break existing setup of ADB version prior to 1.7

praveenkumar:
We have to add in document because since this is first time ADB going to use `openshift` with systemctl. We can create a dist which can be pointed to doc stating if you are using adb-box older than 1.7 then use this vagrant file.

LalatenduMohanty:
@praveenkumar , May be we should have the old copy of the Vagrantfile in the same directory for some time with appropriate name along with some documentation. 
Thoughts @bexelbie @navidshaikh

navidshaikh:
@LalatenduMohanty : I suggest we do not keep the copies of Vagrantfile for a provider. If ADB 1.7 has openshift provided as systemd service, we should ask users to update the box.

We are working on few enhancements in service-manager plugin which should make the enabling / starting a provider in the box with just one line in Vagrantfile.

LalatenduMohanty:
The other way would be creating a branch for every release and point to the branch in in the readme. Similar to what Kubernetes github project does.

navidshaikh:
@LalatenduMohanty : 

> The other way would be creating a branch for every release and point to the branch in in the readme. Similar to what Kubernetes github project does.

+1

LalatenduMohanty:
@navidshaikh, ideally yes. But I dont want to miss on users who might continue using 1.6.0 and somehow they need the old Vagrantfile.

components/centos/centos-openshift-setup/Vagrantfile
```diff
@@ -5,29 +5,18 @@
 DOCKER_REGISTRY="docker.io"
 
 # The name of the OpenShift image.
-ORIGIN_IMAGE_NAME="openshift/origin:v1.1.1"
+ORIGIN_IMAGE_NAME="openshift/origin"
```
bexelbie, line 8: Do we need to allow this value here if we hard code origin for the sub-components below?

Alternately, lines 56-59 should use this tag.

bexelbie:
@LalatenduMohanty 

> The other way would be creating a branch for every release and point to the branch in in the readme. Similar to what Kubernetes github project does.

The users can just go to the tag for the release - there is no need to create branches, imho.

LalatenduMohanty:
@bexelbie , I could not find a way to view the repository for a particular tag using the web browser for the github project. I might not be aware of the method too. Can you point me to somewhere as an example

components/centos/centos-openshift-setup/Vagrantfile
LalatenduMohanty, line 8: @bexelbie , I think we should fix that.

bexelbie:
@LalatenduMohanty got to the branch switcher and click the tiny word tags next to branches below the input box.
ala https://github.com/projectatomic/adb-atomic-developer-bundle/tree/v1.6.0

LalatenduMohanty:
@bexelbie Cool, works for me

LalatenduMohanty:
LGTM

PR #232, opened 2016-02-12
Updates usage docs with renamed plugin vagrant-service-manager

navidshaikh:


bexelbie:
This PR needs to update all docs to reflect new usage. If we don't explain this in other docs, we need to add links.

LalatenduMohanty:
LGTM.

PR #258, opened 2016-02-25
OpenShift Vagrantfile mod to reflect new command

bexelbie:
Do not merge without
https://github.com/projectatomic/adb-utils/pull/45
- sccli now used instead of systemd directly
- docker pull now done in sccli to ensure images for version
- requested are available if run separately
- ORIGIN_\* renamed as not all versions of OpenShift are named ORIGIN
- Environment variables passed to ensure that version is easily
  changable

### DIFF
--- a/components/centos/centos-openshift-setup/Vagrantfile
+++ b/components/centos/centos-openshift-setup/Vagrantfile
@@ -5,8 +5,8 @@
 DOCKER_REGISTRY="docker.io"
 
 # The name of the OpenShift image.
-ORIGIN_IMAGE_NAME="openshift/origin"
-ORIGIN_TAG="v1.1.1"
+IMAGE_NAME="openshift/origin"
+IMAGE_TAG="v1.1.1"
 
 # The public IP address the VM created by Vagrant will get.
 # You will use this IP address to connect to OpenShift web console.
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
@bexelbie Thanks for the PR, I will wait till adb-utils is ready then merge this PR

LalatenduMohanty:
@bexelbie @praveenkumar We can also modify the Vagrantfile to take environment variable i.e. k8s or openshift . That way we can almost unify the Vagrantfile

bexelbie:
@LalatenduMohanty I agree in principal.  I'd prefer to get this pair of patches in separately before working on that.  That way we can fix this to work logically and then focus on the combination.

We should also update vagrant-service-manager to use sccli, imho - @navidshaikh

bexelbie:
@LalatenduMohanty I believe this is ready.

LalatenduMohanty:
@bexelbie Technically yes. But we don't create specific labels in the documentation for each release. That means there is a high probability that an user might use this Vagrantfile with version 1.7.0 and it will fail. 
So we have two issues to fix. 
1. How do we redirect the use to relevant tag version of documents and Vagrantfiles : We can follow Kubernetes project
2. Till we have not fixed the above, we have to wait till we release ADB with the recent adb-utils in it (with your PR) then merge this patch.

bexelbie:
@LalatenduMohanty Can you elaborate on your proposal for following kubernetes?  Today we don't publish our docs onto another site.  We could easily just add a header to the doc files that says this is for latest and you should see the release tag for your version.  Is that sufficient?

I think having to hold PRs for other events is setting us up for wasted effort on PR management.

LalatenduMohanty:
> We could easily just add a header to the doc files that says this is for latest and you should see the release tag for your version. Is that sufficient?

Yes, that should be ok. 

> I think having to hold PRs for other events is setting us up for wasted effort on PR management.

We should not break an user experience in favor of a better PR management.

praveenkumar:
@LalatenduMohanty If we are doing ADB build today/tomorrow then we will be in good position to merge this PR and then me or @bexelbie can work on k8s part.

PR #260, opened 2016-02-25
Deleting unneeded landrush information

bexelbie:
xip.io is currently how OpenShift is provisioned


bexelbie:
@praveenkumar please review

this will prevent confusion as following these directions won't actually change OpenShift

praveenkumar:
LGTM, we will put this back once landrush in mainstream.

PR #262, opened 2016-02-27
Removing the sed commands because these are not required anymore

LalatenduMohanty:
As we are overwriting the eth0 interface in the kickstartfile


bexelbie:
LGTM, without test.  @LalatenduMohanty can you brew a box?

navidshaikh:
LGTM.

LalatenduMohanty:
@bexelbie here is the build https://cbs.centos.org/koji/taskinfo?taskID=78884

PR #298, opened 2016-03-22
Changes for release 1.8.0

LalatenduMohanty:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
 
+## v1.8.0 Mar 23, 2016
+- Changes for release 1.8.0 @LalatenduMohanty
+- OpenShift Vagrantfile mod to reflect new command @bexelbie
+
 ## v1.7.1 Mar 9, 2016
 - Revert "Adding openshift2nulecule to ADB" @LalatenduMohanty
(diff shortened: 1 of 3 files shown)


bexelbie:
Should " +- Update with new meeting time, effective 2016-03-21 @bexelbie " be part of the Changelog?  I believe mentioning it in the release announcement is good, but I always thought our changelog was for code related changes.

LalatenduMohanty:
@bexelbie  I am not sure. Does that mean we should not keep documentation changes part of changelog?

bexelbie:
@LalatenduMohanty I do not consider the change of time of an IRC meeting to be a documentation change.  The meeting time does not help you use the software.  Not knowing hte meeting time doesn't hurt you when using the software.  The meeting time is a community function, not a documentation or code change.

As an example, I am sure the Linux kernel devs have at least one regular meeting.  I suspect I won't find notes about changes in their changelog.

LalatenduMohanty:
@bexelbie  I think you are right. Will fix the changelog

LalatenduMohanty:
@navidshaikh @bexelbie @praveenkumar I think it is ready to be merged.

bexelbie:
LGTM

Issue #274, opened 2016-03-04
Docker registry (distribution) in ADB


tnozicka:
I am convinced we need to have (at least) an optional docker registry in ADB.

In ADB, there is actually installed package _docker-registry-0.9.1-7.el7.x86_64_ but the service is not run or configured in any of our boxes. (We should probably get rid of it since it is the deprecated one.)

That leaves us with 2 options:
1. Use centos' rpm _docker-distribution-2.1.1-3.el7.x86_64_ - which is quite old and will always be in the boxes even if someone opts it out.
2. create an atomic app (nulecule) from docker registry (https://hub.docker.com/_/registry/)

I am inclined to option 2. and I will try to prepare the nuleculized version as part of me getting to understand nulecule properly and getting up to speed.

---

Issue #261 might just mean the same thing but _Atomic registry_ probably means the OpenShift version.

bexelbie:
+1 atomic app
If we don't need docker-distribution in the box, let's remove it from the kickstart

LalatenduMohanty:
@bexelbie @tnozicka  The docker registry was included as we thought a developer might want to have a local registry , but looks like we do not need it. Lets send a mail to container-tools and on lazy consensus we can remove it from ks file

praveenkumar:
FYI, if a Dev is using openshift then local registry is configured and ready to consumed from host machine. Also we are in progress to make it as secure registry.

->  https://github.com/projectatomic/adb-utils/blob/master/services/openshift/scripts/openshift_provision#L44-L49

tnozicka:
@praveenkumar I agree, this is not needed for openshift vagrant box. But there are still 3 ADB boxes (docker, kubernetes, mesos+marathon) which are missing docker registry.

@LalatenduMohanty I think we can safely remove the docker-registry from kickstart files. User can always run it as a docker container and the version we are providing is long deprecated. I will send the email to container-tools list to achieve consensus.

I have been working on AtomicApp for Docker Registry v2 which could be (optionally) used instead.
https://github.com/projectatomic/nulecule-library/pull/71

praveenkumar:
>  But there are still 3 ADB boxes (docker, kubernetes, mesos+marathon) which are missing docker registry.

Yes I agree and we will remove deprecated package from kickstart.

LalatenduMohanty:
Closing the issue as the PR to remove  docker-registry is merged.

PR #300, opened 2016-03-22
Remove docker-registry from installed packages

praveenkumar:
#274  Better to remove deprecated version of docker-registry and then we will see what way we should handle this issue.

### DIFF
--- a/build_tools/kickstarts/centos-7-adb-vagrant.ks
+++ b/build_tools/kickstarts/centos-7-adb-vagrant.ks
@@ -40,7 +40,6 @@ flannel
 bash-completion
 man-pages
 atomic
-docker-registry
 nfs-utils
 PyYAML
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
We should wait few more days before merging this patch

bexelbie:
@LalatenduMohanty why the delay?  We should put the reason in the PR so others can see.

@praveenkumar do you have a scratch build?

LalatenduMohanty:
@bexelbie I thought it is understood because there is a corresponding mail in container-tools and we should give people time to see, read the mail and respond if necessary.

LalatenduMohanty:
Looks like there is no objection with respect to removing the docker-registry from kickstart file [1]

[1] https://www.redhat.com/archives/container-tools/2016-March/msg00170.html

PR #303, opened 2016-03-22
Add cockpit to the ADB

### LINKED ISSUE #228: ADB should include cockpit
We should include cockpit bridge and cache the cockpit-ws interface.  Once ready, service-manager should be extended to include management of cockpit-ws

bexelbie:
Fixes #228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Fix #332 Vagrantfile var name fix @optak
 - Removed kube-apiserver service file duplication @praveenkumar
 - Added Notes on Default for ADB/CDK; Docker version mismatch; and moved relevant VSM steps.@preeticp
+- Fix #228 Add Cockpit @bexelbie
 
 ## v1.8.0 Apr 5, 2016
 - Update Vagrantfiles with RHEL files @LalatenduMohanty
--- a/docs/cockpit.rst
+++ b/docs/cockpit.rst
@@ -1,0 +1,51 @@
+==================================================
+Using the Cockpit with the Atomic Developer Bundle
+==================================================
+
+`Cockpit`_ makes it easy to administer your GNU/Linux servers via a
+web browser.  It can manage system resources and has special functionality
+built in for managing docker containers.  Specifically for containers,
+it can:
+
+* Display a list of images
+* Display a list of containers (started and stopped)
+* Start and stop containers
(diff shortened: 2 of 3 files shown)


bexelbie:
This PR contains the first draft of docs and a fix to add cockpit.

However, the user experience is horrible, in small part due to some issues with cockpit.  This is here while that gets sorted out.

Scratch build: https://cbs.centos.org/koji/taskinfo?taskID=82666

bexelbie:
@LalatenduMohanty I have updated this to reflect my new found knowledge that we provide default passwords (Thank you @rtnpro)

Please review.

A scratch build is running at https://cbs.centos.org/koji/taskinfo?taskID=89066

docs/cockpit.rst
```diff
@@ -0,0 +1,51 @@
...
+
+1. Start the ADB
+
+2. Log into the ADB
+
+   ``$ vagrant ssh``
+
+3. Start cockpit
+
+   ``$ sudo systemctl start cockpit``
```
LalatenduMohanty, line 32: @bexelbie  You are not doing "systemctl enable docker" in the kickstart file because we do not know how may user would actually use Cockpit? or something else?

LalatenduMohanty:
@bexelbie Please rebase the PR.

docs/cockpit.rst
bexelbie, line 32: I am not enabling it by default because I think we need a bit of discussion before we make this a default service.  If we are going to do that we should consider enabling vagrant service-manager.

LalatenduMohanty, line 32: :+1:

hferentschik:
So will cockpit then be available for the CDK as well?

LalatenduMohanty:
@hferentschik Not sure , we have to figure that out.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Fix #332 Vagrantfile var name fix @optak
 - Removed kube-apiserver service file duplication @praveenkumar
 - Added Notes on Default for ADB/CDK; Docker version mismatch; and moved relevant VSM steps.@preeticp
+- Fix #228 Add Cockpit @bexelbie
```
LalatenduMohanty, line 22: @bexelbie The change log got added to 2.0.0 release instead of 2.1.0

Issue #256, opened 2016-02-24
Add check for vagrant-registration plugin


hferentschik:
The default _Vagrantfile_ contains a check whether the _vagrant-service-manager_ plugin is installed, but none for _vagrant-registration_. Given that the config uses _config.registration.username_ and _config.registration.password_ it makes sense to add this test as well.

hferentschik:
Also it makes sense to test for the existence of all required plugins and then report the all missing ones only once. See https://github.com/redhat-developer-tooling/openshift-vagrant/issues/66

hferentschik:
Could we add label focus:devex?

hferentschik:
Since I cannot create a pull request for this. Here is how I think the default Vagrantfile should look like for the CDK atm. Note, the _systemctl enable openshift_ and _systemctl start openshift_ should be removed once OpenShift starts per default - https://gist.github.com/hferentschik/3daaa9297a90016810f4

hferentschik:
PR - https://github.com/projectatomic/adb-atomic-developer-bundle/pull/311

Issue #255, opened 2016-02-24
Allow to configure number of cores and memory size via environment variables


hferentschik:
One could make it easier for users to change the default CPU and memory settings via environment variables. Something like that:

```
# The amount of memory available to the VM
VM_MEMORY = ENV['VM_MEMORY'] || 4096

# Number of VM CPUs
VM_CPU = ENV['VM_CPU'] || 2
```

and then:

```
  cdk.vm.provider "virtualbox" do |v, override|
    override.vm.box = "cdk_v2"
    override.vm.box_url = "#{VAGRANT_VIRTUALBOX_URL}"
    v.name = "openshift.cdk-2"
    v.memory = VM_MEMORY
    v.cpus   = VM_CPU
    v.customize ["modifyvm", :id, "--ioapic", "on"]
    v.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
  end
```

Just as an idea.

praveenkumar:
+1, you want this to be in our default vagrant file?

navidshaikh:
Should the environment variables be prepended by characters specific to our tooling (ADB / CDK) ?

Going forward if we have multiple environment variables, its easier to distinguish them and avoid possible collision with other tooling consuming such generic environment variables names.

For eg:

```
ADB_VM_MEMORY
ADB_VM_CPU

CDK_VM_MEMORY
CDK_VM_CPU
```

We can support for types of env variables which are prepended by either ADB or CDK based on the box in action.

praveenkumar:
> Should the environment variables be prepended by characters specific to our tooling (ADB / CDK) ?

@navidshaikh No, these are common env. variables and should not be different for ADB/CDK or you have any specific use case except basic functionality those variable going to provide? 

> Going forward if we have multiple environment variables, its easier to distinguish them and avoid possible collision with other tooling consuming such generic environment variables names.

Agree for those kind of variable we will identify and add prefix accordingly like we did for Variant file.

hferentschik:
> you want this to be in our default vagrant file?

That was the idea yes. Not a must, but might be nice. If this would go into the box'es _Vagrantfile_ I would definitely say let's do it. While we still ship a separate standalone _Vagrantfile_ it clashes with the goal to keep it as small and simple as possible.

I leave it up to you guys. Maybe we add it and then take the first opportunity we get to move it into the box. BTW, is there any progress on tweaking the ImageFactory build to take a custom _Vagrantfile_?

[...]

Issue #253, opened 2016-02-24
Default Vagrantfile should allow for subscription details via env variables and manual user interaction


hferentschik:
Currently the _Vagrantfile_ contains:

```
  config.registration.username = ENV['SUB_USERNAME']
  config.registration.password = ENV['SUB_PASSWORD']
```

This will allow to get the credentials via environment variables, but it will fail if no env settings are provided. 

Using

```
  cdk.registration.username    = ENV.has_key?('SUB_USERNAME') ? "#{ENV['SUB_USERNAME']}" : nil
  cdk.registration.password    = ENV.has_key?('SUB_PASSWORD') ? "#{ENV['SUB_PASSWORD']}" : nil
```

should bypass this and in case of no env variables, use manual user interaction. 

On this note, it would make sense to add something like this as well, since registration is not needed in these cases:

```
      case ARGV[0]
      when "up", "halt"
        if cdk.registration.username.nil? || cdk.registration.password.nil?
          puts SUBSCRIPTION_INFO
        end
      end
```

hferentschik:
Could you please add label focus:devex

rct:
Note: There are [instructions in the CDK install guide](https://access.redhat.com/documentation/en/red-hat-enterprise-linux-atomic-host/version-7/container-development-kit-installation-guide/#registering_your_red_hat_enterprise_linux_vagrant_boxes_with_red_hat) for putting the registration details in the User's vagrantfile in their home directory. (~/.vagrant.d/Vagrantfile)

```
Vagrant.configure('2') do |config|
 config.registration.subscriber_username = '<your Red Hat username>'
 config.registration.subscriber_password = '<your Red Hat password>'
end
```

So any added check shouldn't overwrite that because the environment variables aren't set (at least without correcting the documentation).

Personally I prefer the user-wide Vagrantfile approach to environment variables. Though environment variables should be able to override ~/.vagrant.d/Vagrantfile as needed.

hferentschik:
A couple of things here. First, the docs are wrong. It uses wrong properties:

```
Vagrant.configure('2') do |config|
 config.registration.subscriber_username = '<your Red Hat username>'
 config.registration.subscriber_password = '<your Red Hat password>'
```

   end

There is no _subscriber_ in the properties names.

[...]

PR #311, opened 2016-04-01
Issue #253, Issue #255, Issue #256 - Improving default OSE Vagrantfile

hferentschik:
- Wrapping the access of cdk.registration.username and cdk.registration.password into an if statement checking for env variables
- Making sure the installation of vagrant-registration is verified
- Allowing to configure VM cores and RAM via env variables

### DIFF
--- a/components/rhel/rhel-ose/Vagrantfile
+++ b/components/rhel/rhel-ose/Vagrantfile
@@ -1,65 +1,68 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-# Usage:
-#
-# To use this Vagrantfile, you(optional):
-# * Connected to internal Red Hat network if you need to any package update
-# * Have 'vagrant-registration' plugin installed
-# * Have valid RH employee subscription account
-#
-# The public IP address the VM created by Vagrant will get.
-# You will use this IP address to connect to OpenShift web console.
+# The private network IP of the VM. You will use this IP to connect to OpenShift.
 PUBLIC_ADDRESS="10.1.2.2"
 
-#check if vagrant-service-manager is installed on system
-unless Vagrant.has_plugin?("vagrant-service-manager")
-  raise "vagrant-service-manager plugin is not installed, run `vagrant plugin install vagrant-service-manager` to install the plugin."
-end
+# Number of virtualized CPUs
+VM_CPU = ENV['VM_CPU'] || 2
+
+# Amount of available RAM
+VM_MEMORY = ENV['VM_MEMORY'] || 3072
+
+# Validate required plugins
+REQUIRED_PLUGINS = %w(vagrant-service-manager vagrant-registration)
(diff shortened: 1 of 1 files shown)


dkutalek:
Do we really want 2 CPUs and 3GB of memory by default for the machine?
We officially support machines starting with 4GB of memory so 3GB of ram for VM
probably means using swap on host for VM memory on these low end machines
(as opposed to guest swapping in case of 1GB RAM for VM).

hferentschik:
>  Do we really want 2 CPUs and 3GB of memory by default for the machine?

I am not so hung up on the actual values. I do think though 2 cores makes 
sense and 1GB is too little. Let's take 2GB then. More important is to get 
in how it is setup.

> We officially support machines starting with 4GB of memory so 3GB of ram for VM

Who actually decided that. A 4GB machine is hardly workable for local 
development only. And now you want to add a VM and OpenShift? 4GB does just 
make no sense imo.

praveenkumar:
#build

components/rhel/rhel-ose/Vagrantfile
```diff
@@ -1,65 +1,68 @@
...
+  if ENV.has_key?('SUB_USERNAME') && ENV.has_key?('SUB_PASSWORD')
+    cdk.registration.username = ENV['SUB_USERNAME']
+    cdk.registration.password = ENV['SUB_PASSWORD']
+  end
+
   config.vm.provision "shell", inline: <<-SHELL
-    sudo systemctl start openshift
-    sudo systemctl enable openshift
+    systemctl enable openshift
+    systemctl start openshift
```
bexelbie, line 46: I believe lines 45 and 46 should be replaced by 

`/usr/bin/sccli openshift`

@praveenkumar ?

bexelbie:
+1 2GB memory recommendation

Added a note about `sccli`

Pending build and @praveenkumar's response, LGTM

components/rhel/rhel-ose/Vagrantfile
praveenkumar, line 46: @bexelbie This PR for CDK vagrant file changes and I don't want to have `sccli` there because it does again try to pull images (even it is there already as cached) which will increase startup time of service by a min or more. We need a way for sccli to figure out if image already exist then don't try pulling it so we can avoid this extra time.

bexelbie, line 46: @praveenkumar can you update this PR with a link to the open issue to fix that.  That way we have a ping to clean it up later :)

praveenkumar, line 46: @bexelbie we do have a issue filed against it https://github.com/projectatomic/adb-utils/issues/51

LalatenduMohanty:
LGTM.

hferentschik:
thanks :-)

PR #320, opened 2016-04-05
Removing the rhel-7-kubernetes-vagrant.ks file

LalatenduMohanty:
As rhel-7-cdk-vagrant.ks serves the purpose.


LalatenduMohanty:
After this PR merged the directory would have 3 ks files i.e.
- centos-7-adb-vagrant.ks  
- rhel-7-atomic-host-vagrant.ks  
- rhel-7-cdk-vagrant.ks

Which looks correct.

bexelbie:
What is `rhel-7-atomic-host-vagrant.ks` used for?  It doesn't, on first read, seem to build a CDK.

LalatenduMohanty:
@bexelbie that should be for RHEL Atomic host.

bexelbie:
Why is the Kickstart for building RHEL-AH living in the CDK/ADB repo?  Shouldn't that be stored somewhere else?

LalatenduMohanty:
@bexelbie The plan was to make atomic host part of the developer story and ecosystem. But looks like it is not the plan anymore. But I would suggest that we should wait some more time to confirm that.

bexelbie:
Can we get the kickstart file relocated to a sub-directory and/or properly labeled?  I'd like to see us suggest to whomever in RH owns it that it either be integrated or moved.  This doesn't seem like a logical place for it to live.

praveenkumar:
@LalatenduMohanty If you agree with @bexelbie then please do needful, This PR is here for a long time.

PR #328, opened 2016-04-07
Fix #326 (dev part) disk partition fix and changes as per release

### LINKED ISSUE #326: Disk partition for CDK kickstart should be similar to ADB
Source: https://www.redhat.com/archives/container-tools/2016-April/msg00007.html

1) the size of the /  partition seems ridiculously large at 20G (he had about 7G used, so maybe 10?)
2) the docker storage setup is using devicemapper. Would it make more sense and be faster to use one of the other storage options? 
3) Docker data size is kinda small at 6g given the vm has a 40g thinly provisioned disk.

Changes required for cdk kickstart file and should be tested on scratch build before committing to dist-git.

praveenkumar:
- Disk file system changed to xfs and also partitioning.
- Updated fuse-sshfs and openshift2nulecule packages
- Updated docker-selinux version
- Increased sleep time to make sure docker daemon start before pull
- Updated OSE version which available to registry.access.redhat.com

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Fix-256: Add check for vagrant-registration plugin @budhrg
 - Refactored code for checking required plugins for CentOS files @budhrg
 - Update README for RHEL component @praveenkumar
+- Fix #326 (dev part) disk partition fix @praveenkumar
 
 ## v1.8.0 Apr 5, 2016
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
LGTM

LalatenduMohanty:
@praveenkumar please add changelog

PR #331, opened 2016-04-08
Replace RPM commands with YUM in CDK Kickstart file

### LINKED ISSUE #330: Replace RPM commands with YUM in CDK Kickstart file.
We should replace https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/build_tools/kickstarts/rhel-7-cdk-vagrant.ks#L61-62 
with https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/build_tools/kickstarts/centos-7-adb-vagrant.ks#L70-71
As I can see docker-selinux  is a dependency of docker.  and docker-selinux is available in RHEL base.

```
[vagrant@rhel-cdk ~]$ rpm -qR docker
/bin/bash
/bin/sh
/bin/sh
/bin/sh
/bin/sh
config(docker) = 1.9.1-25.el7
device-mapper-libs >= 7:1.02.97
docker-forward-journald = 1.9.1-25.el7
docker-selinux >= 1.9.1-25.el7
--------------------------

```

LalatenduMohanty:
Fixes #330

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Replace RPM commands with YUM in CDK Kickstart file @LalatenduMohanty
 - Fix-256: Add check for vagrant-registration plugin @budhrg
 - Refactored code for checking required plugins for CentOS files @budhrg
 - Update README for RHEL component @praveenkumar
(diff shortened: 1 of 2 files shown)


praveenkumar:
@LalatenduMohanty I did test it and it not going to work because post section happen once vm is created and started, this time yum need rhn-register to perform such operation in case of CDK. Please close this issue if you agree on that otherwise we can have more discussion around it.

LalatenduMohanty:
Closing this as we do not need this anymore

PR #335, opened 2016-04-08
Fix #334: Disables openshift service for k8s Vagrantfile

### LINKED ISSUE #334: CDK k8s Vagrantfile needs to disable default openshift setup
Issue and its explanation is at https://github.com/projectatomic/vagrant-service-manager/issues/162

TLDR:
- VSM plugin can only support to configure `docker` and `openshift` services via Vagrantfile. 
- For `kubernetes` services, one has to explicitly enable it via shell provisioning block
- starting openshift service for CDK box is a default setup in VSM plugin
- Case where user configures `kubernetes` via shell provisioning for CDK box, it conflicts with default setup of `openshift` and `kubernetes` service is not started properly, rather `openshift` service is started

Fix:
- Disable `openshift` service in VSM config in k8s Vagrantfile
- Ideal fix: Have support for configuring `kubernetes` service via the VSM plugin and remove the need of shell provisioning block altogether for kubernetes

### LINKED ISSUE #338: Default OSE Vagrantfile still contains systemctl provisioning steps
The default Vagrantfile still contains:

```
  cdk.vm.provision :shell, inline: <<-SHELL
    systemctl enable openshift
    systemctl start openshift
  SHELL
```

With vagrant-service-manager 1.0.0 Openshift gets automatically enabled and started. The systemctl calls are not needed.

navidshaikh:
Fixes #334
 Fixes #338 

Also updates the CDK openshift Vagrantfile.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix #334: Disables openshift service for CDK k8s Vagrantfile @navidshaikh
 - Fix-256: Add check for vagrant-registration plugin @budhrg
 - Refactored code for checking required plugins for CentOS files @budhrg
 - Update README for RHEL component @praveenkumar
--- a/components/rhel/rhel-ose/Vagrantfile
+++ b/components/rhel/rhel-ose/Vagrantfile
@@ -47,10 +47,8 @@ Vagrant.configure(2) do |config|
     config.registration.password = ENV['SUB_PASSWORD']
   end
 
-  config.vm.provision "shell", inline: <<-SHELL
-    systemctl enable openshift
-    systemctl start openshift
-  SHELL
+  # vagrant-service-manager plugin will configure openshift service by default for CDK
+  # config.servicemanager.services = "docker, openshift"
 
   config.vm.provision "shell", inline: <<-SHELL
     echo
--- a/components/rhel/misc/rhel-k8s-singlenode-setup/Vagrantfile
+++ b/components/rhel/misc/rhel-k8s-singlenode-setup/Vagrantfile
@@ -35,6 +35,9 @@ Vagrant.configure(2) do |config|
   end
   config.vm.network "private_network", type: "dhcp"
 
+  # Fix for ADB #334
+  config.servicemanager.services = 'docker'
+
   config.vm.provision "shell", inline: <<-SHELL
      sudo mkdir -p /etc/pki/kube-apiserver/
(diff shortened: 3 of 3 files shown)


LalatenduMohanty:
@navidshaikh Dont we need the same change for ADB k8s file? I guess not as for CDK openshift is default

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Fix #334: Disables openshift service for k8s Vagrantfile @navidshaikh
```
LalatenduMohanty: s/k8s/CDK k8s/ if it is only applicable to CDK

navidshaikh:
> Dont we need the same change for ADB k8s file?

Default setup:
CDK --> docker and openshift
ADB --> docker

LalatenduMohanty:
> CDK --> docker and openshift

right.

components/rhel/rhel-ose/Vagrantfile
```diff
@@ -47,10 +47,8 @@ Vagrant.configure(2) do |config|
     config.registration.password = ENV['SUB_PASSWORD']
   end
 
-  config.vm.provision "shell", inline: <<-SHELL
-    systemctl enable openshift
-    systemctl start openshift
-  SHELL
+  # vagrant-service-manager plugin will configure openshift service by default for CDK
+  # config.servicemanager.services = "docker, openshift"
```
LalatenduMohanty, line 51: Why not remove it completely? as this is not required.

LalatenduMohanty, line 51: However if we are going with the change , it wont be backward compatible with service-manager < 1.0.0?

navidshaikh, line 51: > However if we are going with the change , it wont be backward compatible with service-manager < 1.0.0?

0.0.5 version plugin had support to configure openshift service by default, however there was a bug https://github.com/projectatomic/vagrant-service-manager/issues/131 , which is fixed in 1.0.0 release of the plugin. 
So 1.0.0 version of the plugin is required.

navidshaikh, line 51: > Why not remove it completely? as this is not required.

This is to give user with some hints about default setup right in Vagrantfile. Since the later block is showing up the connection information (echo msgs) , however apparently there is not configuration of ose service. IMO, we should keep those lines as commented in the Vagrantfile.

LalatenduMohanty:
LGTM

PR #365, opened 2016-04-22
Removing the workaround of downgrading python-docker-py. Fixes #356

### LINKED ISSUE #356: python-docker-py package should be removed from kickstart
Currently we have listed https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/build_tools/kickstarts/centos-7-adb-vagrant.ks#L55-L57 `python-docker-py` package which was added because of openshift2nulecule and now we don't require it any more.

LalatenduMohanty:
As the base distro has python-docker-py-1.7.2 and it works with atomic CLI
and on http://mirror.centos.org/centos-7/7/atomic/x86_64/adb/ we have
python-docker-py-1.7.0.0-1.

Previous discussion https://github.com/projectatomic/adb-atomic-developer-bundle/issues/279#issuecomment-194110557

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- Remove the workaround for downgrading python-docker-py @LalatenduMohanty
 - Update os-release to 2.0.0 @LalatenduMohanty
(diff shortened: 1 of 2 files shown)


navidshaikh:
LGTM

PR #366, opened 2016-04-25
Fix #353 added VSM config and removed Vagrantbox Readme

### LINKED ISSUE #353: Do we need docker-base-setup and Vagrantbox ReadMe?
Currently we have docker-base-setup Vagrantfile (https://github.com/projectatomic/adb-atomic-developer-bundle/tree/master/components/centos/centos-docker-base-setup)  which I am not sure now what is the use-case because vagrant service-manager can do docker provision now. So either we can remove this file or we can add vsm config to configure docker.

Also we have a README file which seems outdated and may be not required in this repo (https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/components/boxes/vagrant-boxes.md)

@LalatenduMohanty @bexelbie thoughts?

praveenkumar:
- Add VSM config setting for docker vagrantfile
- Remove Vagrantbox Readme (not needed anymore).


components/centos/centos-docker-base-setup/Vagrantfile
```diff
@@ -53,4 +53,5 @@ Vagrant.configure(2) do |config|
     config.ssh.private_key_path = ENV['OS_PRIVATE_KEYPATH']
     config.ssh.username = 'vagrant'
   end
+  config.servicemanager.services = 'docker'
```
navidshaikh: VSM configures only `docker` service by default for ADB box, we do not need to explicitly mention it, although we can comment out these lines in Vagrantfile to give hints to users.

praveenkumar:
@navidshaikh Please review, I commented out that line in Vagrantfile to give hints to users.

LalatenduMohanty:
LGTM

PR #375, opened 2016-05-11
Removed development group packages from kickstart

praveenkumar:
@development group packages are around 100MB in size and we don't require all those packages so removing it reduced a bit size for vagrant box.


bexelbie:
@praveenkumar I am very in favor of this.  For the future and for testing purposes can you brew a test box and provide a list of the actual packages that got removed?

LalatenduMohanty:
https://cbs.centos.org/koji/taskinfo?taskID=92820

LalatenduMohanty:
LGTM

PR #386, opened 2016-05-16
Update instructions for cockpit

bexelbie:
- use vagrant-service-manager to start cockpit
- Link cockpit instructions to README - including a slight reorder

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -185,17 +185,19 @@ requirements that are not enabled in those boxes.
 Documentation
 =============
 
-* `Architecture and Roadmap`_
-* `Building the Vagrant box`_ for Developers
 * `Installing the ADB`_
 * `How to use the ADB`_
+** `Using Cockpit with the ADB`_
 * `Updating the ADB`_
+* `Architecture and Roadmap`_
+* `Building the Vagrant box`_ for Developers
 
-.. _Architecture and Roadmap: docs/architecture.rst
-.. _Building the Vagrant box: docs/building.rst
 .. _Installing the ADB: docs/installing.rst
 .. _How to use the ADB: docs/using.rst
+.. _Using Cockpit with the ADB: docs/cockpit.rst
 .. _Updating the ADB: docs/updating.rst
+.. _Architecture and Roadmap: docs/architecture.rst
+.. _Building the Vagrant box: docs/building.rst
 
 Interested in Contributing to this project?
 ===========================================
--- a/docs/cockpit.rst
+++ b/docs/cockpit.rst
@@ -12,7 +12,7 @@ it can:
 * Start and stop containers
 * Pull and delete images
 
-.._Cockpit: http://cockpit-project.org/
+.. _Cockpit: http://cockpit-project.org/
 
 ----------------------------
 Accessing Cockpit in the ADB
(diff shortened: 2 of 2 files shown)


docs/cockpit.rst
```diff
@@ -23,29 +23,21 @@ the host you are running the ADB on.
 
 1. Start the ADB
 
-2. Log into the ADB
+2. Start cockpit
 
-   ``$ vagrant ssh``
+   ``$ vagrant service-manager restart cockpit``
```
LalatenduMohanty, line 28: @bexelbie @budhrg this wont work without code in VSM, right?

brgnepal, line 28: For this, `cockpit` need to be in "start" state by default and this doesn't match with step instruction `start cockpit`. User may get confused why we need to `restart` to start `cockpit`. 

IMO there are two approaches:
- Enable `cockpit` by default. Then this step will gone.
- Don't enable `cockpit` by default. Start it through `service-manager` plugin as `$ vagrant service-manager start cockpit` or through Vagrantfiles.

bexelbie, line 28: My test just now showed it working.  Cockpit was not started or enabled before I ran this command.

I agree that "restart" is a bad verb here, but we don't have "start" yet.

LalatenduMohanty, line 28: I am not in favor of another Vagrantfile for Cockpit, as it does does not fall in the categories of K8s, OpenShift, Mesos etc.

brgnepal, line 28: > I agree that "restart" is a bad verb here, but we don't have "start" yet.

@bexelbie  I can quickly work on adding `start` or `enable` command. It should be trivial.

bexelbie, line 28: @LalatenduMohanty I understand your position on Vagrantfiles.  Can you help me relate your comment to this PR?

bexelbie, line 28: @budhrg I've added https://github.com/projectatomic/vagrant-service-manager/pull/218

It relies on the behavior of systemctl restart

LalatenduMohanty, line 28: @bexelbie I am not sure about the `restart` in `vagrant service-manager restart cockpit` . Basically we are asking user to restart the service even if under the hood it is starting it.

LalatenduMohanty, line 28: But I do not have a better suggestion either at this point. So I am confused.

bexelbie, line 28: @LalatenduMohanty It says 'restart' today because that is the only verb supported in vsm.  @budhrg is working on a patch that will extend my PR VSM/#218 to add 'start' as a verb.  When that merges, this PR will update to read 'start' and not 'restart'.  You made a comment, that seemed accidental, about Vagrantfiles that was confusing.

LalatenduMohanty, line 28: As  we do not have alternative to "restart" verb at this moment and we definitely want to have a doc for Cockpit I am merging this.

README.rst
```diff
@@ -185,17 +185,19 @@ requirements that are not enabled in those boxes.
 Documentation
 =============
 
-* `Architecture and Roadmap`_
-* `Building the Vagrant box`_ for Developers
 * `Installing the ADB`_
 * `How to use the ADB`_
+** `Using Cockpit with the ADB`_
```
LalatenduMohanty, line 190: @bexelbie  It broke the formatting https://github.com/projectatomic/adb-atomic-developer-bundle#documentation

LalatenduMohanty, line 190: @bexelbie Fixed it as part of #388

PR #391, opened 2016-05-18
Updated cockpit guide to start cockpit

brgnepal:
Updated as a result of following PR change in Vagrant service manager plugin.
https://github.com/projectatomic/vagrant-service-manager/pull/219

### DIFF
--- a/docs/cockpit.rst
+++ b/docs/cockpit.rst
@@ -23,19 +23,19 @@ the host you are running the ADB on.
 
 1. Start the ADB
 
-2. Start cockpit
+2. Start Cockpit
 
-   ``$ vagrant service-manager restart cockpit``
+   ``$ vagrant ssh -c "sudo systemctl start cockpit"``
 
 3. Determine the IP Address of the ADB
 
    ``$ vagrant service-manager box ip``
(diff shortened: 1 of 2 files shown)


docs/cockpit.rst
```diff
@@ -23,19 +23,19 @@ the host you are running the ADB on.
 
 1. Start the ADB
 
-2. Start cockpit
+2. Start Cockpit
 
-   ``$ vagrant service-manager restart cockpit``
+   ``$ vagrant service-manager start cockpit``
```
hferentschik: I am not opposed to a `service-manager start/stop`, but still don't like that we use it to control cockpit. For now, I think we would be better of to use `systemctl` for that.

brgnepal:
@praveenkumar @bexelbie @LalatenduMohanty 

Since I have reverted all the changes related to `systemd` services in this PR https://github.com/projectatomic/vagrant-service-manager/pull/219. I would better suggest here to go with `systemctl` approach .

praveenkumar:
@budhrg LGTM

bexelbie:
LGTM

Issue #398, opened 2016-05-20
Unexpected package change in build of ADB


bexelbie:
ADB 2.1 recently shipped.  It contains these docker packages:

```
Installed Packages
docker.x86_64                                     1.9.1-25.el7.centos                    @koji-override-0
docker-forward-journald.x86_64                    1.9.1-25.el7.centos                    @koji-override-0
docker-selinux.x86_64                             1.9.1-25.el7.centos                    @extras         
```

A build of ADB that I submitted today contains these packages:

```
Installed Packages
docker.x86_64                                      1.9.1-40.el7.centos                   @koji-override-0
docker-common.x86_64                               1.9.1-40.el7.centos                   @koji-override-0
docker-forward-journald.x86_64                     1.9.1-40.el7.centos                   @koji-override-0
docker-selinux.x86_64                              1.9.1-40.el7.centos                   @extras         
```

How can I get a build that has what was in ADB 2.1?

@LalatenduMohanty any advice?

LalatenduMohanty:
@bexelbie With the existing build scripts you can not get the exact old RPMs through a build if the CentOS base OS has new packages now.  Because we use CentOS's RPM tree. If the tree has updated RPMs the Vagrantbox will get new RPMs. If you want to get exact RPMs then you have to use `yum downgrade` command to do it either in the kickstartfile or in a Vagrantfile.

bexelbie:
Should we have reproducible builds as part of this project?

praveenkumar:
> Should we have reproducible builds as part of this project?

If we want reproducible builds then we always have to tag version to different packages like docker=1.9.1-25 which I don't think feasible because in case of purge, for older version of packages it might fail.

LalatenduMohanty:
Closing this as I believe this is resolved.

PR #399, opened 2016-05-20
Add cockpit-kubernetes to enable Kube use case

### LINKED ISSUE #396: Do we need Cockpit packages pre-installed in ADB?
As per discussion in https://github.com/projectatomic/adb-atomic-developer-bundle/issues/389 we are not making Cockpit part of default experience in ADB. 

So the question is if we are not making Cockpit part of default experience then is there any value  by having Cockpit packages pre-installed in Cockpit. Because Cockpit is available from CentOS base OS, an user can just run command `yum install Cockpit` to get it installed in addition to the steps we have mentioned i.e. https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/cockpit.rst.
Also removing the packages will reduce the size by 20MB approximately. (will check and confirm)

bexelbie:
The kubernetes cockpit plugin isn't installed by default
This also reduces the code installed by only installing the interface
and the docker, kubernetes, and network components.

Storage is ommitted because we don't need to drag in the dependencies
and the use case doesn't seem strong.

Fix #396

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+-Fix #396 - modify Cockpit to have Kubernetes bridge and not drag in unecessary packages @bexelbie
 
 ## v2.1.0 May 19, 2016
 - Fix #353: Add VSM config to docker setup and removed Vagrantbox Readme @praveenkumar
--- a/build_tools/kickstarts/centos-7-adb-vagrant.ks
+++ b/build_tools/kickstarts/centos-7-adb-vagrant.ks
@@ -48,7 +48,11 @@ adb-utils
 centos-release-adb
 fuse-sshfs
 openshift2nulecule
-cockpit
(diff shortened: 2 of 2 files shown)


bexelbie:
Comments desired.  Also, should we have this auto-start just for our Kube story, per an indirect suggestion by @LalatenduMohanty?

Test box pending because of #398 and #400

LalatenduMohanty:
I would suggest we wait for some more feedback on the related issue i.e. https://github.com/projectatomic/adb-atomic-developer-bundle/issues/396 then decide on this PR.

praveenkumar:
@LalatenduMohanty We already waiting long does we have any concern/issue with it except it will increase some VM size?

LalatenduMohanty:
I am still not convinced that we need to ship Cockpit for Kuebernetes and OpenShift GUI. Because Kubernetes community focusing on http://kubernetes.io/docs/user-guide/ui/ and I do not see Cockpit there.

PR #402, opened 2016-05-23
 Add openshift tag information to kickstart instead in sccli/openshift_options

praveenkumar:
Discussion happened to https://github.com/projectatomic/adb-utils/pull/104

### DIFF
--- a/build_tools/kickstarts/rhel-7-cdk-vagrant.ks
+++ b/build_tools/kickstarts/rhel-7-cdk-vagrant.ks
@@ -163,16 +163,21 @@ sleep 120
 # Modify this as needed based on what you want to pre-pull
 # The tag renaming may not be strictly needed but having an image name
 # that points to an internal hostname will likely confuse people
+OPENSHIFT_TAG="v3.2.0.44"
 
-docker pull openshift3/ose:v3.1.1.6
-docker pull openshift3/ose-haproxy-router:v3.1.1.6
-docker pull openshift3/ose-deployer:v3.1.1.6
-docker pull openshift3/ose-docker-registry:v3.1.1.6
-docker pull openshift3/ose-sti-builder:v3.1.1.6
+docker pull openshift3/ose:$OPENSHIFT_TAG
+docker pull openshift3/ose-haproxy-router:$OPENSHIFT_TAG
+docker pull openshift3/ose-deployer:$OPENSHIFT_TAG
+docker pull openshift3/ose-docker-registry:$OPENSHIFT_TAG
+docker pull openshift3/ose-sti-builder:$OPENSHIFT_TAG
 
 echo "Finished pull, running docker images for log debugging"
 docker images
 kill -9 $DOCKER_PID
+
+# Edit openshift_options to reflect the desired default version of OpenShift
+sed -i "/IMAGE=.*/d" /etc/sysconfig/openshift_option
+echo "IMAGE=\"registry.access.redhat.com/openshift3/ose:${OPENSHIFT_TAG}\"" >> /etc/sysconfig/openshift_option
 
 # Clear yum package and metadata cache
 yum clean all


bexelbie:
LGTM - untested

build_tools/kickstarts/rhel-7-cdk-vagrant.ks
```diff
@@ -198,17 +198,23 @@ sleep 120
...
 
 echo "Finished pull, running docker images for log debugging"
 docker images
 kill -9 $DOCKER_PID
 
+# Edit sccli.py and openshift_options as per openshift_version
+OPENSHIFT_SED_TAG=`echo $OPENSHIFT_TAG | sed -e 's|\/|\\\/|g' -e 's|\.|\\\\.|g'`
+sed -i -e "s|^IMAGE_TAG = \"latest\"|IMAGE_TAG = \"${OPENSHIFT_SED_TAG}\"|" /opt/adb/sccli.py
+sed -i -e "s|:latest|:${OPENSHIFT_SED_TAG}|" /etc/sysconfig/openshift_option
+
```
LalatenduMohanty, line 184: I am not sure if this is the right thing to do. Need more time to find out if we have some alternative.

build_tools/kickstarts/centos-7-adb-vagrant.ks
```diff
@@ -70,6 +70,13 @@ yum install -y docker-selinux
 # set tuned profile to force virtual-guest
 tuned-adm profile virtual-guest
 
+OPENSHIFT_TAG="v1.1.6"
+
+# Edit sccli.py and openshift_options as per openshift_version
+OPENSHIFT_SED_TAG=`echo $OPENSHIFT_TAG | sed -e 's|\/|\\\/|g' -e 's|\.|\\\\.|g'`
+sed -i -e "s|^IMAGE_TAG = \"latest\"|IMAGE_TAG = \"${OPENSHIFT_SED_TAG}\"|" /opt/adb/sccli.py
+sed -i -e "s|:latest|:${OPENSHIFT_SED_TAG}|" /etc/sysconfig/openshift_option
+
```
LalatenduMohanty: We do not need this till we have caching infrastructure in CentOS project.

LalatenduMohanty:
@praveenkumar We do not need to modify openshift_options for ADB yet. So we should remove that. Otherwise LGTM

build_tools/kickstarts/centos-7-adb-vagrant.ks
```diff
@@ -70,6 +70,12 @@ yum install -y docker-selinux
 # set tuned profile to force virtual-guest
 tuned-adm profile virtual-guest
 
+OPENSHIFT_TAG="v1.1.6"
+
+# Edit sccli.py and openshift_options as per openshift_version
```
bexelbie: This comment needs to be updated.  Suggested:

Edit openshift_options to reflect the desired default version of OpenShift

build_tools/kickstarts/rhel-7-cdk-vagrant.ks
```diff
@@ -198,17 +198,22 @@ sleep 120
...
+docker pull openshift3/ose-haproxy-router:$OPENSHIFT_TAG
+docker pull openshift3/ose-deployer:$OPENSHIFT_TAG
+docker pull openshift3/ose-docker-registry:$OPENSHIFT_TAG
+docker pull openshift3/ose-sti-builder:$OPENSHIFT_TAG
 
 echo "Finished pull, running docker images for log debugging"
 docker images
 kill -9 $DOCKER_PID
 
+# Edit sccli.py and openshift_options as per openshift_version
```
bexelbie: Same as comment about same line in CentOS KS file

bexelbie:
> @praveenkumar We do not need to modify openshift_options for ADB yet. So we should remove that. Otherwise LGTM

Even if the new code hasn't landed, I don't think there is any side-effect in landing this today.

bexelbie:
I was thinking about this.  Should we actually have separate variables in `openshift_options` for `DOCKER_REGISTRY` `IMAGE_NAME` and `IMAGE_TAG` so that we can copy the vagrantfiles fine grain control?

praveenkumar:
> I was thinking about this. Should we actually have separate variables in openshift_options for DOCKER_REGISTRY IMAGE_NAME and IMAGE_TAG so that we can copy the vagrantfiles fine grain control?

Yes, In past we had that but I think there was a conversation or agreement that we moved back to current syntax, we can look into it again.

praveenkumar:
> Even if the new code hasn't landed, I don't think there is any side-effect in landing this today.

Agreed. @LalatenduMohanty you have any issue with that?

LalatenduMohanty:
> Even if the new code hasn't landed, I don't think there is any side-effect in landing this today.

We should remove it as it is not required. Anything which is not required should not be part of the source code. In long run it helps to have cleaner code.

bexelbie:
@LalatenduMohanty doesn't not landing this code leave us open to the bug where a different version of openshift is started after using sccli?

praveenkumar:
> doesn't not landing this code leave us open to the bug where a different version of openshift is started after using sccli?

No, for ADB, we have openshift tag available in Vagrant file which will make sure we don't have different version of openshift.

LalatenduMohanty:
@praveenkumar LGTM. Two minor issues. 
1. It needs a rebase
2. Does this PR need two commits?

praveenkumar:
@LalatenduMohanty Done.

PR #411, opened 2016-05-30
Fix 407 Error: dev/null: No such file or directory

praveenkumar:
It will fix typo in vagrant file which was causing issue for kubernetes.


LalatenduMohanty:
LGTM

PR #412, opened 2016-05-30
Update origin version to 1.2.0

praveenkumar:
Update openshift origin tag from 1.1.0 to 1.2.0 (Did a sanity test on older ADB box)

### DIFF
--- a/components/centos/centos-openshift-setup/Vagrantfile
+++ b/components/centos/centos-openshift-setup/Vagrantfile
@@ -4,9 +4,10 @@
 # The Docker registry from where we pull the OpenShift Docker image
 DOCKER_REGISTRY="docker.io"
 
-# The name of the OpenShift image.
+# The name of the OpenShift image available on dockerhub.
 IMAGE_NAME="openshift/origin"
-IMAGE_TAG="v1.1.1"
+# Tag of the OpenShift image available on dockerhub.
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
@praveenkumar We should merge this PR as soon as ADB Vagrant box get released with adb-utils 1.7.0. As it has the fix for OpenShift registry and router setup i.e. https://github.com/projectatomic/adb-utils/pull/120

praveenkumar:
@LalatenduMohanty if that looks good to you, it's ready to merge.

LalatenduMohanty:
@praveenkumar will it work fine with ADB 2.1.0 box too without https://github.com/projectatomic/adb-utils/pull/120 ?

praveenkumar:
> will it work fine with ADB 2.1.0 box too without projectatomic/adb-utils#120 ?

No it will not work so adb-utils-1.7 should be required.

components/centos/centos-openshift-setup/Vagrantfile
```diff
@@ -6,7 +6,7 @@ DOCKER_REGISTRY="docker.io"
 
 # The name of the OpenShift image.
```
LalatenduMohanty: @praveenkumar can you please change the comment to make it explicit that we are talking about Docker hub tags here?

Issue #336, opened 2016-04-08
Default to mounting the user's home directory into the box


dustymabe:
A few of us discussed today the fact that we should default to mounting the user's home directory into the vagrant box via sshfs. I'll let this be a tracker for the work that is being done around that.

The idea is that we want to use the `$HOME` variable on Linux/Mac and the `$USERPROFILE` variable on windows to determine where the user's home directory is and then mount that into a predictable location within the guest.

We may need to do windows->linux path conversion for this. Standby for more information on that.

dustymabe:
@hferentschik @LalatenduMohanty - We'll use this as a tracker for the work.

dustymabe:
For sshfs, Hardy is also recommending that we add some extra options to default to open permissions on everything. Something like this should do the trick: 

```
sshfs_opts_append: '-o umask=000 -o uid=1000 -o gid=1000'
```

LalatenduMohanty:
Here are the @hferentschik suggestions I think

```
cdk.vm.synced_folder ENV['HOME'], ENV['HOME'], type: 'sshfs', sshfs_opts_append: '-o umask=000 -o uid=1000 -o gid=1000'   For Linux


cdk.vm.synced_folder ENV['HOME'], ENV['USERPROFILE'], type: 'sshfs', sshfs_opts_append: '-o umask=000 -o uid=1000 -o gid=1000' FOr Windows
```

dustymabe:
Not exactly. One Correction:

```
cdk.vm.synced_folder ENV['USERPROFILE'], ENV['USERPROFILE'], type: 'sshfs', sshfs_opts_append: '-o umask=000 -o uid=1000 -o gid=1000' FOr Windows
```

dustymabe:
It appears using "virtualbox" synced folders does not do windows path conversion. See attached output: [vboxsf.txt](https://github.com/projectatomic/adb-atomic-developer-bundle/files/211464/vboxsf.txt)

Looks like we will need to do it in the Vagrantfile for now. As an alternative we could submit patches back up to Vagrant to do this in each synced folder implementation.

hferentschik:
> It appears using "virtualbox" synced folders does not do windows path conversion. 

:-( Seems wrong to me

> Looks like we will need to do it in the Vagrantfile for now.

I'll look for a way of doing this...

dustymabe:
Thanks @hferentschik

hferentschik:
[...]

Issue #272, opened 2016-03-02
Disable the synchronisation of /vagrant 


hferentschik:
Currently the box Vagrantfile contains:

```
Vagrant.configure("2") do |config|
  config.vm.base_mac = "52540013d96e"
 config.vm.synced_folder ".", "/vagrant", type: "rsync"
end
```

where the sync type for _/vagrant_ is explicitly set to rsync. This is potentially a problem on Windows where rsync might not be available. Also the user might prefer to install vbguest additions. Even with vbguest additions installed he still would be forced to use _rsync_ though. 

That said, I am wondering how much sense it makes to this type of sync anyways. The directory containing the _Vagrantfile_ is most likely irrelevant for the user. There is no need to sync it at all. 

Also, we are investigating options (sshfs or vbguest additons) to mount the whole user directory into the VM (to allow Docker file mounting). In this case it becomes even less relevant to sync the Vagrant working directory to the VM. 

Why do we not just disable the sync:

```
config.vm.synced_folder '.', '/vagrant', disabled: true
```

It gives the user more flexibility regarding the sync options and it also makes setup on Windows easier.

Thoughts?

hferentschik:
Could we add label focus:devex as well, please.

bexelbie:
@hferentschik My understanding is that once we have an alternative file sync in place this will be disabled. Until then we need something.

LalatenduMohanty:
@hferentschik Agree. I have sent a PR to disable it for Virtualbox (was thinking of Windows + Virtualbox) https://github.com/redhat-imaging/imagefactory/pull/367

For others, rsync makes sense. However the settings in Vagrantfile ( which the user uses e.g. the one created by `vagrant init` ) takes precedence, so have not disabled them by default. Let me know if you differ.

hferentschik:
But it is not needed. At least not for syncing /vagrant. Why would we like to sync this directory?

LalatenduMohanty:
@hferentschik @bexelbie Also the issue should live in https://github.com/redhat-imaging/imagefactory

hferentschik:
> @hferentschik Agree. I have sent a PR to disable it for Virtualbox redhat-imaging/imagefactory#367

nice. Need to be verified on Windows. My guess is that it was only added in the first place, since Vagrant tries to use vbguest additions as default on Windows otherwise. With completely disabling it, it should not be an problem though.

> For others, rsync makes sense. 

[...]

PR #413, opened 2016-05-30
Issue #336 and #272 - User home dir mount via vagrant-sshfs

hferentschik:


components/rhel/rhel-ose/Vagrantfile
```diff
@@ -47,6 +47,7 @@ Vagrant.configure(2) do |config|
     config.registration.password = ENV['SUB_PASSWORD']
   end
 
+  config.vm.synced_folder '.', '/vagrant', disabled: true
```
hferentschik, line 50: I am disabling the /vagrant folder sync here, since I don't know what to do to disable it in the default _Vagrantfile_ w/i the box itself. Who knows?

hferentschik:
@praveenkumar any idea how we can make sure that this gets properly tests on the various targeted OS types, in particular Windows 7 and 10?

praveenkumar:
@hferentschik doesn't we require same patch for all Vagrant files (Including CDK/ADB)?

praveenkumar:
> any idea how we can make sure that this gets properly tests on the various targeted OS types, in particular Windows 7 and 10?

QE can test this patch, we should ask them use those Vagrant file on our latest nightly and check if everything as expected. We don't have any automation tests in place for this feature.

hferentschik:
> We don't have any automation tests in place for this feature.

openshift-vagrant does have a sanity check. It verifies there is something mounted. However, we need to also test that working with the actual file system works. If nothing else we need some manual tests to start with.

hferentschik:
> doesn't we require same patch for all Vagrant files (Including CDK/ADB).

most likely. With "all" you mean all Vagrantfiles under the _components_ sub-directory, right?

praveenkumar:
> With "all" you mean all Vagrantfiles under the components sub-directory, right?

yes, that's right.

hferentschik:
> > With "all" you mean all Vagrantfiles under the components sub-directory, right?
> 
> yes, that's right.

How do we go about the /vagrant folder sync? Should I disable it in these Vagrantfiles as well? Or maybe just add the vagrant-sshfs config?

praveenkumar:
> How do we go about the /vagrant folder sync? Should I disable it in these Vagrantfiles as well? Or maybe just add the vagrant-sshfs config?

If we just add vagrant sshfs config does it work with `rsync` and `sshfs`? if yes then we can keep it otherwise we just disable it.

hferentschik:
> If we just add vagrant sshfs config does it work with rsync and sshfs? if yes then we can keep it otherwise we just disable it.

It works. The two issues are orthogonal, but I want to make sure we keep the /vagrant sync in mind and work on it asap. Disabling it in the default Vagrantfile would be much better, since it would automatically affect all usages.

praveenkumar:
> Disabling it in the default Vagrantfile would be much better, since it would automatically affect all usages. 

Agreed, I think we had long discussion around it and can revisit #272 but that's something we have to do from imagefactory side iirc.

hferentschik:
So I would vote for adding vagrant-sshfs to all _Vagrantfiles_ (as per pull request) and then disable the /vagrant sync for CDK. Thoughts?

praveenkumar:
> So I would vote for adding vagrant-sshfs to all Vagrantfiles (as per pull request) and then disable the /vagrant sync for CDK. Thoughts?

+1, lets do that.

hferentschik:
@praveenkumar what's the right place to document this?

praveenkumar:
@hferentschik from ADB side I think best place to have it documented https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/using.rst#vagrant-bi-directional-folder-sync and from CDK side we can ask downstream folks to do it at proper location.

praveenkumar:
Tested with linux and working as expected.

LalatenduMohanty:
#build

hferentschik:
ok, I did some basic doc updates in _using.rst_. I think it would be nice to also then document how this can be leveraged for Docker volume mounts, but using.rst does not seem to be the right place for this. Thoughts?

LalatenduMohanty:
@hferentschik I can see the file permissions changes in side the Vagrant box. I am using Fedora. Do you know why?

```
on my laptop
--------------

[lmohanty@LalatenduM-laptop ~]$ pwd
/home/lmohanty
[lmohanty@LalatenduM-laptop ~]$ ls -l
total 164
.....................................

-rw-rw-r--.   1 lmohanty lmohanty     0 Oct 16  2015 version.hwm
-rw-rw-r--.   1 lmohanty lmohanty     0 Oct 16  2015 version.pwd
-rw-rw-r--.   1 lmohanty lmohanty     0 Oct 16  2015 version.pwi
drwxr-xr-x.   2 lmohanty lmohanty  4096 Jun 10  2015 Videos


Inside the Vagrant box
......................

-rwxrwxrwx. 1 vagrant vagrant     0 Oct 15  2015 version.hwm
-rwxrwxrwx. 1 vagrant vagrant     0 Oct 15  2015 version.pwd
-rwxrwxrwx. 1 vagrant vagrant     0 Oct 15  2015 version.pwi
-drwxrwxrwx. 1 vagrant vagrant  4096 Jun 10  2015 Videos
```

LalatenduMohanty:
> document how this can be leveraged for Docker volume mounts, but using.rst does not seem to be the right place for this. Thoughts?

Agree, We should not just restrict our self at the existing separation if docs e.g. using.rst, If we want to scale. So we should add new documents when required and link it to places applicable.

praveenkumar:
Some more observations. If we create file in the vagrant box then permission appear with `root` group.

```
[HOST]$ touch hello.txt
[centos@hp-dl380pgen8-07 ~]$ ls -l
total 0
-rw-rw-r--. 1 centos centos 0 Jun  9 05:17 hello.txt

[HOST]# vagrant ssh
Last login: Thu Jun  9 05:12:46 2016 from 192.168.121.1
[vagrant]$ cd /home/centos/
[vagrant$ ls -l
total 0
-rwxrwxrwx. 1 vagrant vagrant 0 Jun  9 05:17 hello.txt

[vagrant]$ touch hel.txt
[vagrant]$ ls -l
total 0
-rwxrwxrwx. 1 vagrant vagrant 0 Jun  9 05:17 hello.txt
-rwxrwxrwx. 1 vagrant vagrant 0 Jun  9 05:18 hel.txt
[vagrant]$ exit
logout
Connection to 192.168.121.80 closed.

[HOST]$ ls -l
total 0
-rw-rw-r--. 1 centos centos 0 Jun  9 05:17 hello.txt
-rw-r--r--. 1 root   root   0 Jun  9 05:18 hel.txt
```

praveenkumar:
> I can see the file permissions changes in side the Vagrant box. I am using Fedora. Do you know why?

Because of `sshfs_opts_append: '-o umask=000 -o uid=1000 -o gid=1000'`  set `umask=000` so if you remove it from your vagrant file and have only `sshfs_opts_append: '-o uid=1000 -o gid=1000'` then this issue will not come up. @hardy we want to know why we are having `umask=000` option in our Vagrant file any specific reason?

> If we create file in the vagrant box then permission appear with root group.

I was getting it because I was running vagrant box using `root` so that is not the issue.

LalatenduMohanty:
LGTM

hferentschik:
> We should not just restrict our self at the existing separation if docs e.g. using.rst, If we want to scale. So we should add new documents when required and link it to places applicable.

+1, however for my part I wait for asciidoctor conversion. I am just more used to it and I am not a big fan of the rst syntax. Also I cannot find proper tooling support, aka a decent preview while editing.

hferentschik:
> we want to know why we are having umask=000 option in our Vagrant file any specific reason

Jupp, you want to make sure the files are word read-/writ-able on the guest. Otherwise you are running into file permissions problems when you then try to use this files/directories in Docker volume mounts. If the user/group ids don't match between VM and container, you are having issues. BTW, this is also what docker-machine does.

hferentschik:
Thanks for merging.

LalatenduMohanty:
@hferentschik The only issue is if user by mistakenly deletes home that would be an issue. in case of docker-machine it is difficult to do as user does not login to the VM. But for CDK/ADB `vagrant ssh` gives full access.

PR #416, opened 2016-05-30
Fix #414 kube-apiservice shouldn't be part of kickstart file

### LINKED ISSUE #414: Re-examine kube-apiserver service file to be part of CDK kickstart file
Currently we have  https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/build_tools/kickstarts/rhel-7-cdk-vagrant.ks#L69-L99 as part of kickstart file and with current available version of kube rpm we really don't require it any more and this is already fixed for ADB box. Refer: #318

praveenkumar:
With latest version of kube packages we don't really need those steps in our kickstart file.


LalatenduMohanty:
LGTM

PR #417, opened 2016-05-30
Fix #415 add yum clean all step to ks file to remove yum local cache

### LINKED ISSUE #415: Add yum clean all step to kickstart file to remove yum local cache
We have to add `yum clean all` at the end of our kickstart file to make sure yum cached will be cleared and it will help us to shrink our vagrant boxes (save ~12 MB).

praveenkumar:
We have to clean yum local cache which save us some space. (~ 10 MB)


LalatenduMohanty:
LGTM

Issue #176, opened 2015-12-20
Re-evaluate packages for the ADB box


LalatenduMohanty:
I think we need to rethink if we need below packages for ADB because these packages does not fit to the objective of the project at this point.

```
@development
flannel
```

I dont think we need @development YUM group. Removing this will reduce the size of the Vagrantbox. Also flannel is required for overlay networking when we have Kubernetes cluster

praveenkumar:
@LalatenduMohanty Let we make a scratch build after removing those and test if that works for our use-case as expected. We really need to shrink vagrant box size.

tkdchen:
How about install packages during provision while `vagrant up`?
Can the build make a minimal installation with Minimal iso image?

bexelbie:
+1 to testing without those.  I am concerned that flannel could be an upstream use case for Kube though

@tkdchen A design goal of the ADB is for it to be, as much as possible, a single download.  We are going to cache some docker images for the same reason.  Your suggestion is valid, but creates a long start up time, and a lot of network traffic if a cache isn't used.

langdon:
Why not file a separate issue to create a 'simple vagrant box' as something available alongside the ADB for more generic use cases. This would also encourage people to file issues about other use cases we might want to have vagrant box support for. This issue can then remain focused on the package set required to make a developer immediately productive with containers.

bexelbie:
Can we start by comparing the packages in a default ADB with an Atomic Host?

See also #190

praveenkumar:
- [x] Check packages which can be removed.
- [x] yum clean all after provision.
- [x] Remove docs from `/usr/share/doc`
- [ ] Look out if any other better way?

LalatenduMohanty:
Closing the issue as this is mostly fixed now.

PR #420, opened 2016-06-02
Attempt to reduce vagrant box size to address #176

praveenkumar:
This PR is consist of experiment around reducing the VM size and incorporate those to make an attempt of having less size of vagrant box (check #176) .

### DIFF
--- a/build_tools/kickstarts/rhel-7-cdk-vagrant.ks
+++ b/build_tools/kickstarts/rhel-7-cdk-vagrant.ks
@@ -26,7 +26,7 @@ logvol / --size=8192 --fstype="xfs" --name=root --vgname=VolGroup00
 
 reboot
 
-%packages
+%packages  --excludedocs --instLangs=en --nocore
 docker
 deltarpm
 rsync
@@ -37,11 +37,9 @@ screen
 git
 kubernetes
 etcd
-flannel
 bash-completion
 man-pages
 atomic
-docker-registry
 PyYAML
 #libyaml-devel
 tuned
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
@praveenkumar LGTM, have you tested the scratch build?

praveenkumar:
> have you tested the scratch build?

@LalatenduMohanty for CDK yes but ADB I didn't created a scratch build, can you do that and let me know, will test it out.

LalatenduMohanty:
Scratch build https://cbs.centos.org/koji/taskinfo?taskID=94839

dustymabe:
Don't know if this works in CentOS 7 version of anaconda but in fedora you can specify to `excludedocs` from the `%packages` statement: 
https://pagure.io/fedora-kickstarts/blob/master/f/fedora-docker-base.ks#_31

dustymabe:
you could also look at reducing the set of supported languages in the image as was also done in that same kickstart file (YMMV with respect to this working in older versions of Anaconda though).

praveenkumar:
@dustymabe @LalatenduMohanty Thanks guys I am able to reduce around 50 MB more after reducing set of supported languages.

build_tools/kickstarts/rhel-7-cdk-vagrant.ks
```diff
@@ -58,6 +84,9 @@ openshift2nulecule
 # Workaround for BZ1336857
 restorecon -v /usr/bin/docker*
 
+LANG="en_US"
+echo "%_install_lang $LANG" > /etc/rpm/macros.image-language-conf
```
LalatenduMohanty, line 88: I do not understand how this works. @praveenkumar @dustymabe do you guys have some pointers for me?

praveenkumar, line 88: As per doc http://rpm.org/api/4.4.2.2/config_macros.html, since we only care about en_US then package installed should have only en_US locale instead of other languages.

```
A colon separated list of desired locales to be installed;
"all" means install all locale specific files.

%_install_langs all
```

LalatenduMohanty, line 88: @praveenkumar thanks

LalatenduMohanty:
LGTM

Issue #117, opened 2015-11-03
SELinux vs Kubernetes volumes 


kadel:
SELinux inside ADB is blocking using emptyDir and hostPath volumes and passing secrets to containers as volumes
#### emptyDir and hostPath  volumes

If you create pod with volume, where volume is either emptyDir or hostPath,
SELinux is blocking access to that directory inside container.
##### example

busybox.yaml:

```
apiVersion: v1
kind: Pod
metadata:
  name: busybox
  namespace: default
spec:
  containers:
  - image: busybox
    name: busybox
    command:
      - sleep
      - "36000"
    volumeMounts:
      - name: storage
        mountPath: /storage
  volumes:
    - name: storage
      hostPath:
        path: /tmp/storage 
```

```
kubectl create -f busybox.yaml
```

```
kubectl exec -it busybox /bin/sh
/ # touch /storage/asdf
touch: /storage/asdf: Permission denied
```

I have to manually change context of /tmp/storage on host to get that working

```
chcon -Rt svirt_sandbox_file_t /tmp/storage
```

Same issue is with emptyDir volumes. This situation is a little bit complicated because emptyDir volumes are created in path containing pod uid  `/var/lib/kubelet/pods/4dd7b77b-8228-11e5-b8db-525400e09276/volumes/kubernetes.io~empty-dir/`
#### Kubernetes secrets

SELinux is also blocking access to secret volumes so you can not pass secrets from kubernetes to containers.
##### example:

secret.yaml

```
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
type: Opaque
data:
  password: dmFsdWUtMg0K
  username: dmFsdWUtMQ0K
```

busybox.yaml

```
apiVersion: v1
kind: Pod
metadata:
  name: busybox
  namespace: default
spec:
  containers:
  - image: busybox
    name: busybox
    command:
      - sleep
      - "36000"
    volumeMounts:
      - name: storage
        mountPath: /storage
      - name: secret
        mountPath: /secret

  volumes:
    - name: storage
      hostPath:
        path: /tmp/storage 
    - name: secret
      secret:
        secretName: mysecret
```

```
kubectl  create  -f secret.yaml
kubectl  create  -f busybox.yaml
```

now you should be able to access `/secret/username` and `/secret/password` inside container

```
# kubectl exec -it busybox /bin/sh
# ls /secret/
ls: /secret/username: Permission denied
ls: /secret/password: Permission denied
```

[...]

PR #425, opened 2016-06-07
Fix #289 Can not change securityContext in k8s

### LINKED ISSUE #289: Can not change securityContext in k8s
Let's say you want to run some image in k8s as different user than the preset one in the image. If you are using docker, you can change user by '-u' option. Kubernetes has securityContext.runAsUser for this case but in ADB it is disabled by apiserver's configuration option **SecurityContextDeny**.

``` bash
KUBE_ADMISSION_CONTROL="--admission_control=NamespaceLifecycle,NamespaceExists,LimitRanger,SecurityContextDeny,ServiceAccount,ResourceQuota"
```

IMHO this is wrong and we should remove the option **SecurityContextDeny**.

---

One use case is mounting volumes in nuleculized application which should be able to use different images. Every image can have different users (root, 42, ...) but the owner of the mounted directory is still the same. It is solved by using securityContext.runAsUser.

praveenkumar:
This patch will fix securityContext for k8s and also address #117 (except documentation part).

### DIFF
--- a/build_tools/kickstarts/centos-7-adb-vagrant.ks
+++ b/build_tools/kickstarts/centos-7-adb-vagrant.ks
@@ -87,6 +87,10 @@ restorecon -v /usr/bin/docker*
 echo "dockerlog:x:4294967295:4294967295::/var/lib/docker:/bin/nologin" >> /etc/passwd
 echo "dockerlog:x:4294967295:4294967295::/var/lib/docker:/bin/nologin" >> /etc/group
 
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
@praveenkumar you mean we need documentation also for this?

praveenkumar:
@LalatenduMohanty yes for #117, if user want to set selinux policy for their container and using k8s.

LalatenduMohanty:
LGTM.

PR #426, opened 2016-06-07
Update variant information

praveenkumar:
This patch will include updated variant information for CDK.


LalatenduMohanty:
LGTM

build_tools/kickstarts/rhel-7-cdk-vagrant.ks
```diff
@@ -95,7 +95,7 @@ echo "%_install_lang $LANG" > /etc/rpm/macros.image-language-conf
 # https://github.com/projectatomic/adb-atomic-developer-bundle/issues/183
 echo "VARIANT=\"Container Development Kit (CDK)\"" >> /etc/os-release
 echo "VARIANT_ID=\"cdk\"" >> /etc/os-release
-echo "VARIANT_VERSION=\"2.0\"" >> /etc/os-release
+echo "VARIANT_VERSION=\"2.1\"" >> /etc/os-release
```
hferentschik, line 98: BTW, I think this hard coded version string is not enough. You really want to know the exact build. This needs to be somehow passed in dynamically as part of the build process. I had several cases where I trouble shooted VM issues and it turned out in the end that we had different versions of the VM (this is in particular important now that we have weekly and even daily builds)

PR #432, opened 2016-06-09
Fix # 408 ADB Installation Doc modification

Preeticp:
Added vagrant service manager plugin installation step. Modified the installation document for better flow. Added vagrantfiles and steps to use them for better user experience.

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -11,25 +11,27 @@ Installing the Atomic Developer Bundle
 1. Install a Virtualization Provider
 ------------------------------------
 
-Two virtualization providers have been tested with the ADB.
+Two virtualization providers have been tested with the ADB; VirtualBox and Libvirt.
+
+The following virtualization providers are suggested for the respective operating systems:
 
 * Microsoft Windows and Mac OS X
 
-  The suggested virtualization provider is `VirtualBox`_. Installation
+  The suggested virtualization provider for Microsoft Windows and Mac OS X is `VirtualBox`_. Installation
   instructions are available `online`_. While the latest stable shipping release
   should work, the majority of testing has been done with version 5.0.0 on Mac
   OS X and 5.0.8 on Microsoft Windows.
 
 .. _VirtualBox: https://www.virtualbox.org
-.. _oneline: https://www.virtualbox.org/manual/UserManual.html
+.. _online: https://www.virtualbox.org/manual/UserManual.html
 
 * Fedora
 
-  Two different virtualization providers are supported on Linux, `VirtualBox`_
+  Two different virtualization providers are supported on Linux; `VirtualBox`_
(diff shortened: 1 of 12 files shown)


docs/installing.rst
```diff
@@ -11,25 +11,27 @@ Installing the Atomic Developer Bundle
 1. Install a Virtualization Provider
```
brgnepal, line 11: Could we fix double bullets issues as like below if we click on view button:

double_bullets

Preeticp, line 11: Hi Budh,
In this case they are sequential steps hence numbered. Have you found this issue in any other place too? I will look into this.

brgnepal, line 11: I mean to say why those bullets before 1, 2, 3 etc

docs/installing.rst
```diff
@@ -11,25 +11,27 @@ Installing the Atomic Developer Bundle
...
-  Two different virtualization providers are supported on Linux, `VirtualBox`_
+  Two different virtualization providers are supported on Linux; `VirtualBox`_
   and `libvirt <http://libvirt.org/>`_. The choice as to which to use should be
   driven by your preferences and environmental concerns and is outside of the
   scope of this document. Both will work equally well in their default
-  configuration. You may wish to read the section on file synchronization when
+  configuration. You may wish to read the section on `file synchronization`_ when
   making this decision.
 
   * VirtualBox installation instructions are available `online at the VirtualBox
```
bexelbie: Ugh, we need directions here.  This is the fedora section, so we should be able to advise using the distribution build via:

`sudo dnf install VirtualBox-4.3`

docs/installing.rst
```diff
@@ -120,51 +126,91 @@ Two virtualization providers have been tested with the ADB.
...
+* Vagrant Initiated Download
 
-  The image is available at https://atlas.hashicorp.com/projectatomic/boxes/adb.
   The ``vagrant`` program is capable of downloading the box the first time it is
   needed. This happens when you first initialize a new vagrant environment.
 
-  If you wish to use a project provided vagrant file you should first get the
-  Vagrantfile as directed in `Using the Atomic Developer Bundle`_ in
-  the *Using Custom Vagrantfiles for Specific Use Cases* section.
+  The ADB project provides a few customised Vagrantfiles, which can auto download the ADB, in the `Using Custom Vagrantfiles for Specific Use Cases`_ section of the `Using the Atomic Developer Bundle`_ document. The ADB images are cataloged at `atlas.hashicorp.com`_.
```
bexelbie: The ADB project provides customized Vagrantfiles, which will download the ADB and automatically set up specific container development environments.  They are listed below and more details are available, in the `Using Custom Vagrantfiles for Specific Use Cases`_ section of the `Using the Atomic Developer Bundle`_ document.

praveenkumar:
LGTM.

PR #434, opened 2016-06-10
Updating README_CDK_ZIP and vm name to cdk2.1 in Vagrantfiles

LalatenduMohanty:

### DIFF
--- a/components/rhel/README_CDK_ZIP.rst
+++ b/components/rhel/README_CDK_ZIP.rst
@@ -15,6 +15,8 @@ Here is the mapping of upstream projects and READMEs for CDK.zip file shipped wi
  ├── plugins
  │   ├── vagrant-registration.gem
  │   ├── vagrant-registration-README.md
+ │   ├── vagrant-sshfs.gem
+ │   ├── vagrant-sshfs-README.md
  │   ├── vagrant-service-manager.gem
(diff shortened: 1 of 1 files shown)


praveenkumar:
We should wait before merge this because it will affect our documentations also.

bexelbie:
@praveenkumar can you elaborate?

LalatenduMohanty:
@praveenkumar @bexelbie this is ready to be merged now. I have removed the vmname change from the PR

components/rhel/README_CDK_ZIP.rst
```diff
@@ -15,6 +15,8 @@ Here is the mapping of upstream projects and READMEs for CDK.zip file shipped wi
  ├── plugins
  │   ├── vagrant-registration.gem
  │   ├── vagrant-registration-README.md
+ │   ├── vagrant-sshfs-1.1.0.gem
```
praveenkumar: Let's remove the version info like you are doing for `vagrant-registration` because this file only to map what available in the zip otherwise we have to create PR everytime when there is version update.

LalatenduMohanty: Agree

bexelbie:
LGTM - manually merged to avoid merge commit.

praveenkumar:
LGTM

PR #475, opened 2016-07-18
marathon: setup so that ports from host can be shared

dustymabe:
We are configuring so that ports from the host can be mapped directly
to containers as described in the following link:

http://container-solutions.com/configuring-resource-offers-on-mesos/


LalatenduMohanty:
LGTM. @dharmit comments?

dharmit:
@LalatenduMohanty looks good!

PR #483, opened 2016-07-25
Added dependencies installation for F24

thatdocslady:
Resolves issue #450 

This PR adds a step to the installation instructions for F24, and is set up in a format that allows for future OS-specific prerequisites to be added.

In addition to this PR, there is also a PR in the vagrant-service-manager repo with a prerequisites step that links back to the ADB install instructions (in order to avoid documenting the information in multiple places): https://github.com/projectatomic/vagrant-service-manager/pull/355

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -48,7 +48,7 @@ The following virtualization providers are suggested for the respective operatin
     While the latest stable release should work, the majority of testing has
     been done with version 4.3.30.
 
-  * Installing libvirt dependencies can be skipped as they are automatically installed together with the ``vagrant-libvirt`` package. 
+  * Installing libvirt dependencies can be skipped as they are automatically installed together with the ``vagrant-libvirt`` package.
     Testing has been done with libvirt version 1.2.18 and vagrant-libvirt
     version 0.0.32.
 
@@ -126,7 +126,7 @@ The following virtualization providers are suggested for the respective operatin
       # Set libvirtd to start automatically on system boot
       $ sudo systemctl enable libvirtd
 
-    This would install both Vagrant and Libvirt. 
+    This would install both Vagrant and Libvirt.
 
 * CentOS
 
@@ -139,7 +139,7 @@ The following virtualization providers are suggested for the respective operatin
     $ sudo yum -y install centos-release-scl
     $ sudo yum -y install sclo-vagrant1
     $ sudo scl enable sclo-vagrant1 bash
-  
+
   To add `libvirt` support use this::
 
     # Start libvirtd
@@ -148,15 +148,32 @@ The following virtualization providers are suggested for the respective operatin
     # Set libvirtd to start automatically on system boot
     $ sudo systemctl enable libvirtd
 
------------------------------------------------------
-3. Install the vagrant-service-manager Vagrant plugin
------------------------------------------------------
+----------------------------------
+3. Install additional dependencies
+----------------------------------
 
-The `vagrant-service-manager`_ plugin can be installed using::
+For some operating systems, you might need to install additional dependencies before you install the Vagrant plugins.
 
-    vagrant plugin install vagrant-service-manager
+* Fedora 24
+
+  Run the following commands to install the additional dependencies::
+
+  $ sudo dnf install @'Development Tools'
+  $ sudo dnf install rpm-build zlib-devel ruby-devel gcc-c++
+
+--------------------------
+4. Install Vagrant plugins
+--------------------------
+
+Run the following commands to install the `vagrant-service-manager`_, `vagrant-sshfs`_, and `landrush`_ plugins::
+
+$ vagrant plugin install vagrant-service-manager
+$ vagrant plugin install vagrant-sshfs
+$ vagrant plugin install landrush
 
 .. _vagrant-service-manager: https://github.com/projectatomic/vagrant-service-manager
+.. _vagrant-sshfs: https://github.com/dustymabe/vagrant-sshfs
+.. _landrush: https://github.com/vagrant-landrush/landrush
 
 -------------------
 4. Download the ADB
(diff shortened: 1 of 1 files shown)


docs/installing.rst
```diff
@@ -148,8 +148,22 @@ The following virtualization providers are suggested for the respective operatin
     # Set libvirtd to start automatically on system boot
     $ sudo systemctl enable libvirtd
 
+---------------------------------------------------
+3. Install dependencies for vagrant-service-manager
```
bexelbie: Should we document these dependencies in the VSM repo and reference them from here?  I think we have to document these dependencies in VSM regardless.

LalatenduMohanty: @bexelbie agree

LalatenduMohanty: @bexelbie I mean keeping the information at one place will help maintaining it better.  We should document it here and reference it in VSM because the user facing docs should be here as ADB acts as landing page for the ecosystem projects.

docs/installing.rst
```diff
@@ -148,8 +148,22 @@ The following virtualization providers are suggested for the respective operatin
...
+
+  Run the following commands to install ``Development Tools``, ``rpm-build``, and ``zlib-devel`` packages::
+
+  $ sudo dnf install "Development Tools"
+  $ sudo dnf install rpm-build
+  $ sudo dnf install zlib-devel
+
 -----------------------------------------------------
-3. Install the vagrant-service-manager Vagrant plugin
+4. Install the vagrant-service-manager Vagrant plugin
```
bexelbie: We now require vagrant-sshfs as well, I believe it is appropriate to add that in this PR

thatdocslady:
@bexelbie @LalatenduMohanty I agree with both of you on the need to have the info available in VSM too, which is why I also added the prerequisite step in the VSM install instructions (see link to PR in the VSM repo). This way we document the details here and reference it from there.

I considered handling it the other way around (documenting there and referencing here), but then realized that the VSM install instructions are conceptually nested in the global ADB instructions, and new users are more likely to follow the ADB docs than the VSM docs.

docs/installing.rst
thatdocslady: @bexelbie who can I talk to about the vagrant-sshfs instruction? I'd like to make sure that we cover whatever is needed for this install step.

Preeticp:
Should these dependencies be added as a part of the higher level step 'Install the vagrant-service-manager Vagrant plugin', to keep it consistent with the rest of the doc?
All the previous steps seem to have variations dealing with respective Operating Systems as bullet points within the higher level step.

thatdocslady:
@Preeticp in this specific case the dependencies are required not just for VSM but also for sshfs, so it makes sense to sign-post them independently. However, you raise a good point and we should discuss how to standardize prerequisites across the board. If you don't mind, maybe it's good to open a separate issue of type enhancement with some examples of usage and a call for discussion. Thanks!

docs/installing.rst
hferentschik: I'd not list the dependencies per plugin. I'd rather list everything one needs to install on Fedroa to get going in one go.

And  I am against listing all these detailed instructions on vagrant-serice-manager. The service-manager docs already point to ADB docs for more detailed help. I think we can keep it this way.

docs/installing.rst
```diff
@@ -148,8 +148,22 @@ The following virtualization providers are suggested for the respective operatin
...
+
+For some operating systems, you might need to install additional dependencies.
+
+* Fedora 24
+
+  Run the following commands to install ``Development Tools``, ``rpm-build``, and ``zlib-devel`` packages::
+
+  $ sudo dnf install "Development Tools"
+  $ sudo dnf install rpm-build
+  $ sudo dnf install zlib-devel
```
hferentschik: I believe you are missing ruby-devel:

```
 sudo dnf install -y vagrant vagrant-libvirt
 sudo dnf install -y @"Development Tools" rpm-build gcc-c++
 sudo dnf install -y ruby-devel zlib-devel
```

Who is verifying the required commands set?

docs/installing.rst
hferentschik: I think similar to the comment about listing each and every dependency per plugin, we should just have a single install Vagrant plugin install section. It could be as easy as:

```
# CDK only
$ vagrant plugin install vagrant-registration
$ vagrant plugin install landrush
$ vagrant plugin install vagrant-sshfs
$ vagrant plugin install vagrant-service-manager
$
```

docs/installing.rst
hferentschik: BTW, I thought the plan was to convert these documents to asccidoc?

thatdocslady:
@hferentschik I requested verification in the issue, not sure who else should be looking at these.

docs/installing.rst
```diff
@@ -148,15 +148,36 @@ The following virtualization providers are suggested for the respective operatin
...
+----------------------------------
+
+For some operating systems, you might need to install additional dependencies before you install the Vagrang plugins.
+
+* Fedora 24
+
+  Run the following commands to install the additional dependencies::
+
+  $ sudo dnf install "Development Tools"
+  $ sudo dnf install rpm-build
```
bexelbie: Can we group the dnf commands into one line?

docs/installing.rst
```diff
@@ -148,15 +148,36 @@ The following virtualization providers are suggested for the respective operatin
...
+
+--------------------------
+4. Install Vagrant plugins
+--------------------------
 
-The `vagrant-service-manager`_ plugin can be installed using::
+Run the following commands to install the required Vagrant plugins::
 
-    vagrant plugin install vagrant-service-manager
+$ vagrant plugin install vagrant-registration
```
bexelbie: AFAIK, vagrant-registration is not required for the ADB.  Since CDK docs are currently separate, I believe it should be deleted.

@LalatenduMohanty @hferentschik

LalatenduMohanty: Yes, vagrant-registration is only required for CDK

thatdocslady:
@LalatenduMohanty @bexelbie updated the PR with latest feedback, back to you :)

docs/installing.rst
```diff
@@ -148,13 +148,30 @@ The following virtualization providers are suggested for the respective operatin
...
+3. Install additional dependencies
+----------------------------------
+
+For some operating systems, you might need to install additional dependencies before you install the Vagrant plugins.
+
+* Fedora 24
+
+  Run the following commands to install the additional dependencies::
+
+  $ sudo dnf install "Development Tools"
```
LalatenduMohanty: This is not the correct command 

```
$ sudo dnf install "Development Tools"
Last metadata expiration check: 0:24:11 ago on Tue Aug  9 11:34:14 2016.
No package Development Tools available.
Error: Unable to find a match.
```

Instead please use `sudo dnf install @'Development Tools'`

```
[vagrant@localhost ~]$ sudo dnf install @'Development Tools'
Last metadata expiration check: 0:25:35 ago on Tue Aug  9 11:34:14 2016.
Dependencies resolved.
===================================================================================================================================================================================================================
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

Transaction Summary
===================================================================================================================================================================================================================
Install  72 Packages
Upgrade   1 Package

Total download size: 83 M
Is this ok [y/N]: 
```

docs/installing.rst
```diff
@@ -148,13 +148,30 @@ The following virtualization providers are suggested for the respective operatin
...
+  $ sudo dnf install rpm-build zlib-devel ruby-devel
```
LalatenduMohanty: Found another dependency for vagrant plugins in Fedora 24 i.e. `sudo dnf install gcc-c++` . It provides g++ compiler and without it landrush did not get installed.

LalatenduMohanty:
@thatdocslady Can you please squash the commits in to one. Sorry missed commenting about it last time. I think the commits are logically part of a single commit.

LalatenduMohanty:
LGTM

docs/installing.rst
```diff
@@ -148,13 +148,30 @@ The following virtualization providers are suggested for the respective operatin
...
 
-The `vagrant-service-manager`_ plugin can be installed using::
+Run the following commands to install the required Vagrant plugins::
 
-    vagrant plugin install vagrant-service-manager
+$ vagrant plugin install landrush
+$ vagrant plugin install vagrant-sshfs
+$ vagrant plugin install vagrant-service-manager
+
+See the `vagrant-service-manager`_ for more information about this plugins.
```
bexelbie: This needs to be expanded to provide links for all three plugin repos

PR #487, opened 2016-08-03
Updated latest openshift tag

praveenkumar:
Updated to latest tag which is available at registry.access.redhat.com


LalatenduMohanty:
LGTM

PR #490, opened 2016-08-03
Fix #489: Update usage doc to Document OpenShift configurations

### LINKED ISSUE #489: Update usage doc to Document OpenShift configurations as only works if we mention all three options
Hi,

Due to this [issue](https://github.com/projectatomic/vagrant-service-manager/issues/366), update existing existence to warn users that it will only work if all three options are provided.

brgnepal:
Fix #489

### DIFF
--- a/docs/using.rst
+++ b/docs/using.rst
@@ -127,6 +127,12 @@ To use ADB with Host-Based tools:
    2. ``config.servicemanager.openshift_image_name = "openshift/origin"`` - Specifies the image to be used.
    3. ``config.servicemanager.openshift_image_tag = "v1.2.0"`` - Specifies the version of the image to be used.
 
+   **Note:** Pending resolution of `Issue #366 <https://github.com/projectatomic/vagrant-service-manager/issues/366/>`_ in the vagrant-service-manager plugin, one needs to use all the three variables for any customization.
+
+   Due to  in
+   vagrant-service-manager, one need to use all the
+   three variables for any of the customization.
+
 4. Start the ADB using ``vagrant up``. For details consult the `Installation documentation`_.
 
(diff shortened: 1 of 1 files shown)


praveenkumar:
LGTM, lets discuss with @LalatenduMohanty and @hferentschik

docs/using.rst
```diff
@@ -127,6 +127,10 @@ To use ADB with Host-Based tools:
    2. ``config.servicemanager.openshift_image_name = "openshift/origin"`` - Specifies the image to be used.
    3. ``config.servicemanager.openshift_image_tag = "v1.2.0"`` - Specifies the version of the image to be used.
 
+   **Note:** Due to `this issue <https://github.com/projectatomic/vagrant-service-manager/issues/366/>`_ in
+   vagrant-service-manager, one need to use all the
+   three variables for any of the customization.
+
 4. Start the ADB using ``vagrant up``. For details consult the `Installation documentation`_.
```
Preeticp, line 136: Pending resolution of `Issue #366<https://github.com/projectatomic/vagrant-service-manager/issues/366/>`_ in the vagrant-service-manager plugin, one needs to use all the three variables for any customization.

docs/using.rst
```diff
@@ -127,6 +127,10 @@ To use ADB with Host-Based tools:
    2. ``config.servicemanager.openshift_image_name = "openshift/origin"`` - Specifies the image to be used.
    3. ``config.servicemanager.openshift_image_tag = "v1.2.0"`` - Specifies the version of the image to be used.
 
+   **Note:** Due to `this issue <https://github.com/projectatomic/vagrant-service-manager/issues/366/>`_ in
```
brgnepal: Review comment here.

praveenkumar:
@budhrg This is now not required because we already have fix which is going to part of our all new release.

Issue #488, opened 2016-08-03
Need to integrate vagrant-registration proxy variables with OpenShift proxy variables


LalatenduMohanty:
The current proxy variables in Red Hat/OpenShift Vagrantfile looks as below

https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/components/rhel/rhel-ose/Vagrantfile#L15

```
#PROXY = "<proxy_server>:<port>"
#PROXY_USER = "foo"
#PROXY_PASSWORD = "mysecretpass"
```

And vagrant-registration needs the information as below

```
  # Specify a HTTP proxy to use. This config option if of the format [<host>|<ip>:<port>], eg mongo:8080
  config.registration.proxy

  # Specify a username to use with an authenticated HTTP proxy
  config.registration.proxyUser

  # Specify a password to use with an authenticated HTTP proxy
  config.registration.proxyPassword
```

IMO we should unify both proxy entries in to one set of entries

LalatenduMohanty:
The possible solution I can think of is using ENV['<>']

```

config.registration.proxy = ENV['PROXY']
config.registration.proxyUser = ENV['PROXY_USER']
config.registration.proxyPassword = ENV['PROXY_PASSWORD']
```

```
PROXY = ENV['PROXY'']
PROXY_USER = ENV['PROXY_USER']
PROXY_PASSWORD = ENV['PROXY_PASSWORD']
```

praveenkumar:
Fixed by PR #493

PR #493, opened 2016-08-05
Fixes issue #488. Using ENV variables for proxy

LalatenduMohanty:

### DIFF
--- a/components/rhel/rhel-ose/Vagrantfile
+++ b/components/rhel/rhel-ose/Vagrantfile
@@ -16,12 +16,6 @@ VM_CPU = ENV['VM_CPU'] || 2
 # Amount of available RAM
 VM_MEMORY = ENV['VM_MEMORY'] || 3072
 
-# Variables to enable HTTP/HTTPS proxy
-
-#PROXY = "<proxy_server>:<port>"
-#PROXY_USER = "foo"
-#PROXY_PASSWORD = "mysecretpass"
-
 # Validate required plugins
 REQUIRED_PLUGINS = %w(vagrant-service-manager vagrant-registration vagrant-sshfs landrush)
 errors = []
@@ -36,10 +30,6 @@ unless errors.empty?
   fail Vagrant::Errors::VagrantError.new, msg
 end
 
-# Set default values for proxy variables if not defined.
-PROXY ||= ''
-PROXY_USER ||= ''
-PROXY_PASSWORD ||= ''
 
 Vagrant.configure(2) do |config|
   config.vm.box = 'cdkv2'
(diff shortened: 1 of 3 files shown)


LalatenduMohanty:
@praveenkumar I have tested this and works for me. 
@naina-verma can you please give it a spin on Windows

praveenkumar:
@LalatenduMohanty It mean you don't have any variable in the Vagrantfile where you can let use edit those details and everytime they have to use env variable for proxy?

LalatenduMohanty:
@praveenkumar Basically the idea is to give customer single way to give proxy information to CDK setup i.e. for both vagrant-registration and OpenShift. Also  I think we need to support environment variables. But if we can find someway to also provide the information through Vagrantfile that will be good.

naina-verma:
@LalatenduMohanty  I tested it with hyperv and virtualbox providers , works fine !! ACK

components/rhel/rhel-ose/Vagrantfile
```diff
@@ -64,6 +54,26 @@ Vagrant.configure(2) do |config|
     config.registration.password = ENV['SUB_PASSWORD']
   end
 
+  if ENV.has_key?('PROXY')
```
brgnepal: You can try it in one line as :

```
  config.registration.proxy = PROXY = (ENV['PROXY'] || '')
  # other similar variables
```

In above case, lines from 73 to 75 are not required.

praveenkumar: @budhrg So we can remove L#57-#75 by using below code right?

```
config.registration.proxy = PROXY = (ENV['PROXY'] || '')
config.registration.proxyUser = PROXY_USER = (ENV['PROXY_USER'] || '')
config.registration.proxyPassword = PROXY_PASSWORD = (ENV['PROXY_PASSWORD'] || '')
```

brgnepal: Yes @praveenkumar . Well it should work and better to do it to be on safe side :smile:

LalatenduMohanty:
@budhrg @praveenkumar I tested the new change and it is working fine.

LalatenduMohanty:
I need to add the documentation change too

components/rhel/rhel-ose/Vagrantfile
```diff
@@ -64,6 +54,11 @@ Vagrant.configure(2) do |config|
     config.registration.password = ENV['SUB_PASSWORD']
   end
 
+  #Proxy Information from environment
```
brgnepal: Space after `#` is recommended.

praveenkumar:
LGTM :+1:

PR #496, opened 2016-08-08
Updated kubernetes github link

brgnepal:
Updated kubernetes documentation github link


praveenkumar:
@budhrg we don't need it anymore because that service is already part of kubernetes packaging.

PR #502, opened 2016-08-11
Add Landrush as plugin requirement in Vagrantfile

praveenkumar:
This patch will add landrush as dependent plugin for CentOS now.


bexelbie:
LGTM

PR #504, opened 2016-08-12
Fix #503 Minor change to Vagrantfile to enable oc login

### LINKED ISSUE #503: Unable to do 'oc login openshift.cdk ' from vm
Unable to do oc login openshift.cdk from vm 

Steps to reproduce 
1. Vagrant up with Unified vagrantfile which has landrush settings in it 
2. Do vagrant ssh 
3. oc login openshift.cdk:8443  `error: dial tcp: lookup openshoft.cdk: no such host`
4. After adding openshift.cdk into /etc/hosts file it works 
openshift.cdk is not accessible from inside of vm
Environment : windows 10 pro , hyperv provider

praveenkumar:
Landrush generated TLD is not pingable from guest. If a user do vagrant ssh then it openshift is running and can be accessible by localhost also so no need to mention TLD.

### DIFF
--- a/components/rhel/rhel-ose/Vagrantfile
+++ b/components/rhel/rhel-ose/Vagrantfile
@@ -99,7 +99,7 @@ Vagrant.configure(2) do |config|
     echo
     echo "To use OpenShift CLI, run:"
     echo "$ vagrant ssh"
-    echo "$ oc login #{OPENSHIFT_HOSTNAME }:8443"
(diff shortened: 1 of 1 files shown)


praveenkumar:
@naina-verma  does that look good?

brgnepal:
LGTM

naina-verma:
@praveenkumar , its good !! ACK

PR #507, opened 2016-08-12
Updated centos-k8s Vagrantfile to start through service-manager hook

brgnepal:
This patch is need to run `kubernetes` through vagrant-service-manager as part of `vagrant up` hook so that kube config can be generated properly.

One alternative way is to only generate kube config through vagrant-service-manager plugin only.
Right now, it is getting generate through two ways:
- One during vagrant up process only if kubernetes is mentioned in `config.servicemanager.services` configuration
- If user start kubernetes through CLI as `vagrant service-manager start kubernetes`

Another approach could be:
- Generate when user fire `env kubernetes` command
- Generate when user install kubernetes client binary through `install-cli` command.

### DIFF
--- a/components/centos/centos-k8s-singlenode-setup/Vagrantfile
+++ b/components/centos/centos-k8s-singlenode-setup/Vagrantfile
@@ -64,16 +64,5 @@ Vagrant.configure(2) do |config|
     sudo setsebool -P virt_sandbox_use_fusefs 1
   SHELL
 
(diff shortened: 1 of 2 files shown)


brgnepal:
Just tested with CDK 11th august nightly build. Change working fine for CDK too.

praveenkumar:
LGTM also tested.

Issue #500, opened 2016-08-11
Vagrant failed to initialize at a very early stage , due to version upgrade (Documentation note)


naina-verma:
Vagrant failed to initialize at a very early stage , due to version upgrade (Documentation note)
Hi @Preeticp @praveenkumar ,
Logs link : https://gist.github.com/naina-verma/25110683f570ecdf4921ac8010c89c63

This Error occurred when user upgrade from vagrant older version (1.7.4 ) to (1.8.*) . A doc Note would be good which  point  to remove older meta data. By doing following steps:
1. uninstall vagrant 
2. remove this C:/Users/cdk/.vagrant.d/ directory 
3. install 1.8.1 or 1.8.5

thatdocslady:
Hi @naina-verma I'll be documenting this, can you help by providing some distro/env information? We should estimate whether this is a problem that's relevant only for certain distros or global.

Also, would this be a candidate for a code bug fix, or something that we should accept as a permanent limitation?

Thanks!

naina-verma:
@thatdocslady ,
This problem is on Windows machine , this issue comes into picture when some one upgrade there system form vagrant 1.7.4 version to 1.8.1 or 1.8.\* .
Few of the binaries are not supported for those versions . So few sequential steps are required to achieve a clean up and running  setup;
Those steps are:
1. uninstall vagrant
2. remove this C:/Users/{username}/.vagrant.d/ directory {rm -rf C:/Users/{username}/.vagrant.d/ }
3. install 1.8.1 or 1.8.5

thatdocslady:
@naina-verma @bexelbie can you please review PR #514 :)

LalatenduMohanty:
Ideally this issue should be reported in Upstream Vagrant Or we can create a trouble shooting document to address these kind of issues.

naina-verma:
@thatdocslady , #514  looks good to me !

thatdocslady:
@LalatenduMohanty I understand and agree, but at this point we're on the verge of a full restructuring of the doc set so creating a troubleshooting/known issues document will happen then. In the meantime, this is the best section that we have to hold this information.

LalatenduMohanty:
@thatdocslady In that case if we have an issue to track what we agreed here , it should be fine to put the documentation as per the PR.

thatdocslady:
@LalatenduMohanty created #524 to address this and future "we shouldn't forget to do this" tasks :)

[...]

PR #514, opened 2016-08-19
Added note about updating ADB on Windows machines and format fixes

thatdocslady:
Fixes issue #500 

Added a note about additional steps required on Windows machines to update, and also restructured the Updating topic a little to accommodate for the new content and any future modular additions.

Note: I intentionally didn't list version numbers because that's a maintenance headache, and the only justification for that would be if the limitation is temporary, in which case it doesn't belong in the documentation but instead should be listed in the release notes.

### DIFF
--- a/docs/updating.rst
+++ b/docs/updating.rst
@@ -1,36 +1,63 @@
-===============================================================
-How to update the Atomic Developer Bundle to the latest version
-===============================================================
+===========================================================
+How to update Atomic Developer Bundle to the latest version
+===========================================================
 
-* If you are using the image from atlas.hashicorp.com
+Depending on your implementation of the ADB, you might be able to directly
+update your Vagrant box. In some cases, you will need to perform a clean
+installation or follow additional steps to complete the update.
 
-  1. Check if a newer version of the box is available use this command::
+-------------
+Prerequisites
+-------------
 
-      # vagrant box outdated
+1. If ADB is running, run the `vagrant halt` command::
 
-  2. If a newer version is available, the output should look like this::
+    # vagrant halt
 
-        # vagrant box outdated
+2. Delete the virtual machine with the `vagrant destroy` command. This
+   action will also delete any data or configuration in your ADB that is not saved
+   on your host machine::
 
-     Checking if box 'projectatomic/adb' is up to date...
-     A newer version of the box 'projectatomic/adb-testing' is available! You
-     currently have version '1.3.1'. The latest is version '1.3.2'. Run ``vagrant
-     box update`` to update.
+    # vagrant destroy
 
-  3. If you see the above output, update the ADB Vagrant image with this
-     command::
+-----------------------------------
+Update ADB from atlas.hashicorp.com
+-----------------------------------
 
-      # vagrant box update
+If you use ADB images from atlas.hashicorp.com, you can perform a direct update of
+the Vagrant boxes.
 
-  4. Once the image is updated, you can restart the ADB with the new version
-     with this command::
+1. Run the following command to check if a newer version of the box is available::
 
-      # vagrant up
+    # vagrant box outdated
 
-* If you are using image from cloud.centos.org
+2. If a newer version is available, the output should look like this::
 
-  You will need to monitor the mailing list and other announcement points to
-  determine if a new image is available. If an image is available, download it
-  and install it using the instructions in the `installation documentation
-  <installing.rst>`_.
 
+      Checking if box 'projectatomic/adb' is up to date...
+      A newer version of the box 'projectatomic/adb-testing' is available! You
+      currently have version '1.3.1'. The latest is version '1.3.2'. Run `vagrant
+      box update` to update.
+
+3. If you see the above output, update the ADB Vagrant image with this
+   command::
+
+    # vagrant box update
+
+--------------------------------
+Update ADB from cloud.centos.org
+--------------------------------
+
+If you use ADB images from cloud.centos.org, you must manually remove the
+previously-installed boxes from your host machine and perform a clean installation.
+
+1. Monitor the mailing list and other announcement points to determine if a new image
+   is available.
+
+2. If you are updating ADB on a Windows machine, run the following command to
+   delete the `.vagrant.d` directory::
+
+     # rm -rf C:/Users/<user_name>/.vagrant.d/
+
+3. Download the latest image and perform a clean install as described in
+   the `installation documentation <installing.rst>`_.


docs/updating.rst
```diff
@@ -2,35 +2,56 @@
...
 
-     Checking if box 'projectatomic/adb' is up to date...
-     A newer version of the box 'projectatomic/adb-testing' is available! You
-     currently have version '1.3.1'. The latest is version '1.3.2'. Run ``vagrant
-     box update`` to update.
+    # vagrant box outdated
 
-  3. If you see the above output, update the ADB Vagrant image with this
-     command::
+   Checking if box 'projectatomic/adb' is up to date...
```
brgnepal: @thatdocslady I think the lines from 21 to 24 should be in grey color as it is output of the `vagrant box outdated` command.

bexelbie:
@budhrg why does centos-ci get so chatty?

brgnepal:
@bexelbie Probably @dharmit could answer you well :smile:

docs/updating.rst
```diff
@@ -2,35 +2,56 @@
...
+-------------------------
+
+On some Windows machines, you might need to perform a clean install regardless
+of the source of the image.
+
+1. Uninstall Vagrant.
+
+2. Run the following command to delete the `.vagrant.d` directory::
+
+   # rm -rf C:/Users/<user_name>/.vagrant.d/
```
bexelbie: is `rm` available in all recommended windows shells?

bexelbie:
@mariel, fyi - at this time there are no release notes for ADB - that needs to happen, but until then we need to be conscious of it in docs.  This doc appears to not be affected.

docs/updating.rst
```diff
@@ -2,35 +2,55 @@
...
+2. If a newer version is available, the output should look like this::
 
-     Checking if box 'projectatomic/adb' is up to date...
-     A newer version of the box 'projectatomic/adb-testing' is available! You
-     currently have version '1.3.1'. The latest is version '1.3.2'. Run ``vagrant
-     box update`` to update.
 
-  3. If you see the above output, update the ADB Vagrant image with this
-     command::
+      Checking if box 'projectatomic/adb' is up to date...
```
brgnepal, line 37: :+1:

thatdocslady:
@budhrg good catch about the output formatting, apparently this was already broken before I started working on it. Fixed it and also removed the redundant command, since it's already described in the previous step.

thatdocslady:
@bexelbie so should I list version numbers for this note? or can this step be useful / non-damaging for any Windows update procedures? better safe than sorry when it comes to updates, from my experience...

docs/updating.rst
```diff
@@ -2,35 +2,55 @@
...
+update your Vagrant box. In some cases, you will need to perform a clean
+installation or follow additional steps to complete the update.
 
-  1. Check if a newer version of the box is available use this command::
+--------------------------------------------------
+Update the ADB for images from atlas.hashicorp.com
+--------------------------------------------------
 
-      # vagrant box outdated
+1. Check if a newer version of the box is available use this command::
```
brgnepal: @thatdocslady  I am not able to read this line properly as expected.

Do you think it should be:

```
To check if a new version of the box is available, use this command::
```

thatdocslady: Right, will fix.

brgnepal:
LGTM except one suggestion.

thatdocslady:
@budhrg fixed the phrasing as you suggested

docs/updating.rst
```diff
@@ -2,35 +2,55 @@
...
-     Checking if box 'projectatomic/adb' is up to date...
-     A newer version of the box 'projectatomic/adb-testing' is available! You
-     currently have version '1.3.1'. The latest is version '1.3.2'. Run ``vagrant
-     box update`` to update.
 
-  3. If you see the above output, update the ADB Vagrant image with this
-     command::
+      Checking if box 'projectatomic/adb' is up to date...
+      A newer version of the box 'projectatomic/adb-testing' is available! You
+      currently have version '1.3.1'. The latest is version '1.3.2'. Run ``vagrant
```
brgnepal: @thatdocslady  Since we did formatting here, the quotes like "``" doesn't have any effect. Can we change to single quote?

Rest is looking fine :smile:

thatdocslady: @budhrg done :)

docs/updating.rst
```diff
@@ -2,35 +2,55 @@
...
-     A newer version of the box 'projectatomic/adb-testing' is available! You
-     currently have version '1.3.1'. The latest is version '1.3.2'. Run ``vagrant
-     box update`` to update.
 
-  3. If you see the above output, update the ADB Vagrant image with this
-     command::
+      Checking if box 'projectatomic/adb' is up to date...
+      A newer version of the box 'projectatomic/adb-testing' is available! You
+      currently have version '1.3.1'. The latest is version '1.3.2'. Run `vagrant
+      box update to update.
```
brgnepal: @thatdocslady 
Probably missed ending back tick after box update.
It should be 

```
Run `vagrant box update` to update. 
```

or

```
Run 'vagrant box update' to update.
```

thatdocslady:
@budhrg that's my punishment for trigger-happy committing ;) fixed now

thatdocslady:
@budhrg @bexelbie I think I fixed everything in the feedback, unless there are other comments can someone please merge this?

docs/updating.rst
```diff
@@ -2,35 +2,55 @@
...
 ===============================================================
 
-* If you are using the image from atlas.hashicorp.com
+Depending on your implementation of the ADB, you might be able to directly
+update your Vagrant box. In some cases, you will need to perform a clean
+installation or follow additional steps to complete the update.
 
-  1. Check if a newer version of the box is available use this command::
+--------------------------------------------------
+Update the ADB for images from atlas.hashicorp.com
```
bexelbie: Should this be "Updating"?

It reads funny to me.  Perhaps "Update the ADB using atlas.hashicorp.com" ?

docs/updating.rst
```diff
@@ -2,35 +2,55 @@
...
-      # vagrant box update
+3. If you see the above output, update the ADB Vagrant image with this
+   command::
 
-  4. Once the image is updated, you can restart the ADB with the new version
-     with this command::
+    # vagrant box update
 
-      # vagrant up
+4. Once the image is updated, you can restart the ADB with the new version
```
bexelbie: This needs to be "start" not "restart", I think (see above)

docs/updating.rst
```diff
@@ -2,35 +2,55 @@
 How to update the Atomic Developer Bundle to the latest version
 ===============================================================
 
-* If you are using the image from atlas.hashicorp.com
+Depending on your implementation of the ADB, you might be able to directly
+update your Vagrant box. In some cases, you will need to perform a clean
+installation or follow additional steps to complete the update.
```
bexelbie, line 7: We need a prerequisite statement that says that you must not only `vagrant halt` your existing ADB, but you must actually `vagrant destroy` it.

This will lose any data or configuration in your ADB that is not saved on your host.

@budhrg please confirm.

docs/updating.rst
thatdocslady: @bexelbie I try to avoid gerunds in task titles, but maybe that's a personal/historical style preference.

Maybe "Update ADB from atlas.h..."?

docs/updating.rst
thatdocslady: @bexelbie @budhrg if you're already discussing this prereq, can you clarify whether it's required only for this image source and at which point to run the command? thx

brgnepal: @thatdocslady @bexelbie :
Yes clean installation requires `vagrant destroy` and we should add a warning that this step will potentially remove any data or configuration the user performed with current ADB VM and it is the user's decision whether to update the Vagrant box or not.

docs/updating.rst
```diff
@@ -2,35 +2,55 @@
...
+-----------------------------------------------
+Update the ADB for images from cloud.centos.org
+-----------------------------------------------
 
+Monitor the mailing list and other announcement points to determine if a new
+image is available. If an image is available, download it and perform a clean
+install as described in the `installation documentation <installing.rst>`_.
+
+-------------------------
+Note for Windows machines
```
LalatenduMohanty: This below section is not for ADB but for updating Vagrant version in Windows. I do not think this section should be part of how to update ADB documentation. This kind of documentation should be part of a separate documentation.

docs/updating.rst
thatdocslady: @budhrg so where should I add this step? Is it relevant only for the centos images? As it looks now, the atlas images can be directly updated without a clean install.

thatdocslady:
@bexelbie @budhrg restructured the topic and added the vagrant destroy prerequisites, please review :)

docs/updating.rst
```diff
@@ -1,36 +1,53 @@
...
+installation or follow additional steps to complete the update.
 
-  1. Check if a newer version of the box is available use this command::
+-------------
+Prerequisites
+-------------
 
-      # vagrant box outdated
+If you cannot directly update your Vagrant box, you must perform the following steps
+to remove the previously-installed boxes from your virtual machine.
```
brgnepal: IMO, "previously-installed boxes from your virtual machine." should be "previously-installed boxes from your host machine."

docs/updating.rst
```diff
@@ -1,36 +1,53 @@
...
 
-  2. If a newer version is available, the output should look like this::
+1. If ADB is running, run the `vagrant halt` command.
```
brgnepal: Some indentation issue here 
screenshot from 2016-09-01 20-19-25

brgnepal: Also note that "vagrant halt" is not visible in code block. I believe it should be wrapped inside double back ticks (rst format)

docs/updating.rst
```diff
@@ -1,36 +1,53 @@
...
 
-        # vagrant box outdated
+2. Delete the Vagrant boxes with the `vagrant destroy` command. This action will also
```
brgnepal: "vagrant destroy" will remove "virtual machine". Please update the sentence as "Remove the Virtual machine with 'vagrant destroy' command.

brgnepal: One more step is required to delete the Vagrant box (as in case of `vagrant box update` it will be done automatically) as:

```
vagrant box remove  projectatomic/adb [or box name]
```

docs/updating.rst
```diff
@@ -1,36 +1,53 @@
...
+delete any data or configuration in your ADB that is not saved on your host.
 
-     Checking if box 'projectatomic/adb' is up to date...
-     A newer version of the box 'projectatomic/adb-testing' is available! You
-     currently have version '1.3.1'. The latest is version '1.3.2'. Run ``vagrant
-     box update`` to update.
+3. If you are updating ADB on a Windows machine, run the following command to delete
```
brgnepal: I think we can specify this point as Note. Since it's a special case.

brgnepal:
@thatdocslady I have reviewed from my side.

All good except few suggestions. Please check.

docs/updating.rst
```diff
@@ -1,36 +1,53 @@
...
+update your Vagrant box. In some cases, you will need to perform a clean
+installation or follow additional steps to complete the update.
 
-  1. Check if a newer version of the box is available use this command::
+-------------
+Prerequisites
+-------------
 
-      # vagrant box outdated
+If you cannot directly update your Vagrant box, you must perform the following steps
```
bexelbie: How do I know if I can directly update or not?

docs/updating.rst
bexelbie: +1

docs/updating.rst
bexelbie: I don't think you need this line and I don't think `vagrant box update` does anything similar.  What will happen is that a new box will be added to the ~/.vagrant directory.  Boxes are versioned so they don't clobber each other.

That said, a note that suggests that removing the box with the above command along with a --version string (or something like that check vagrant help) to free up disk space may be useful.  However, I can see omitting this on the presumption that our job isn't to really teach vagrant ... dunno.

docs/updating.rst
bexelbie: At a minimum, we need to call it out above.

docs/updating.rst
brgnepal: Updating the sentence with hyperlink to the section "Update ADB from atlas.hashicorp.com" should work.

docs/updating.rst
brgnepal: Ya agreed that these steps looking like we are teaching vagrant here :wink: .

IMO there are two ways how these boxes are managed by developers usually.
- Automatic download , update:
  
  If we do `vagrant up` on Vagrantfile where box name describe in `config.vm.box` is not present locally, then vagrant will download box from "atlas.hashicorp.com". This will also allow automatic update with `vagrant box update`.
- Download box, add box:
  
  Second way is to download the box like anywhere from `cloud.centos.org`, then add through `vagrant box add <box_name> <local box url / remote url>` command and specify this `box_name` in Vagrantfile and does "vagrant up". In this case, automatic update is not possible. Then updating will need download new updated box, add that box. If same name is to be used then probably --version is need to be used otherwise user will get error as box name already exists. Otherwise, user need to remove the box and then add updated box.

Hope you will get some insights on how users generally use.

thatdocslady:
@bexelbie @budhrg I think I may have been able to address all your points, did some more restructuring and I feel that the flow of the content is much more logical now for each use-case.

Also, unfortunately admonitions (notes, warnings) in RST and ASCIIDoc are intentionally not live-rendered in GitHub[1], and since the Windows machine step must be performed before you download the new image it's probably better to keep it as a step in the task and the fact that we specify Windows machines is enough IMO.

[1] https://github.com/github/markup/issues/68

brgnepal:
@thatdocslady LGTM :+1:

thatdocslady:
@bexelbie can you please review and merge if LGTY? :)

thatdocslady:
@bexelbie took out the prerequisites to their own section based on @LalatenduMohanty 's needed info, please review and merge if ok

docs/updating.rst
```diff
@@ -1,36 +1,63 @@
...
-     box update`` to update.
+    # vagrant destroy
 
-  3. If you see the above output, update the ADB Vagrant image with this
-     command::
+3. If you are updating ADB on a Windows machine, run the following command to
+   delete the `.vagrant.d` directory::
 
-      # vagrant box update
+     # rm -rf C:/Users/<user_name>/.vagrant.d/
```
bexelbie: I'd like to see the response to https://github.com/projectatomic/adb-atomic-developer-bundle/issues/500#issuecomment-244985560 incorporated here - this feels like a cloud.centos edge case

docs/updating.rst
thatdocslady: It looks like Naina confirmed that this step is required for CDK, which means only CentOS images. Will adjust accordingly.

docs/updating.rst
thatdocslady: @bexelbie see https://github.com/projectatomic/adb-atomic-developer-bundle/issues/500#issuecomment-247329072, naina confirmed this is CDK (which means centos), I'll move the step to that task and rebase.

thatdocslady:
@bexelbie @LalatenduMohanty moved the Windows comment to the section for centos images based on Naina's comment. Please review and merge if ok.

docs/updating.rst
```diff
@@ -1,36 +1,63 @@
...
+Update ADB from cloud.centos.org
+--------------------------------
+
+If you use ADB images from cloud.centos.org, you must manually remove the
+previously-installed boxes from your host machine and perform a clean installation.
+
+1. Monitor the mailing list and other announcement points to determine if a new image
+   is available.
+
+2. If you are updating ADB on a Windows machine, run the following command to
```
bexelbie, line 57: I still think this should be in a note that mentions in case you run into trouble.  I don't think this is a required step for every upgrade.

thatdocslady: @bexelbie we had this discussion before already and I explained that if someone needs to do this step it must be in order, and it's better to sign-post it as a preventive rather than a "if you fail, go back and do this". It's not destructive as much as vagrant destroy, and it's better to run it anyway from what I understand.

bexelbie:
LGTM

brgnepal:
+1

Merged. Thanks for the PR.

PR #530, opened 2016-09-02
Modifying the OpenShift vagrant up output to use VSM install-cli feature

LalatenduMohanty:

### DIFF
--- a/components/centos/centos-openshift-setup/Vagrantfile
+++ b/components/centos/centos-openshift-setup/Vagrantfile
@@ -114,9 +114,11 @@ Vagrant.configure(2) do |config|
     echo
     echo "Configured cluster admin user: admin, Password: admin"
     echo
+    echo "To download and install OC binary, run:"
+    echo "vagrant service-manager install-cli openshift"
+    echo
     echo "To use OpenShift CLI, run:"
-    echo "$ vagrant ssh"
-    echo "$ oc login"
+    echo "$ oc login ${OPENSHIFT_HOSTNAME}:8443"
     echo
     echo "To browse the OpenShift API documentation, follow this link:"
     echo "http://openshift3swagger-claytondev.rhcloud.com"
--- a/components/rhel/rhel-ose/Vagrantfile
+++ b/components/rhel/rhel-ose/Vagrantfile
@@ -77,7 +77,7 @@ Vagrant.configure(2) do |config|
 
   config.vm.provision "shell", run: "always", inline: <<-SHELL
     #Get the routable IP address of OpenShift
-    OSIP=`/opt/adb/openshift/get_ip_address`
+    OS_IP=`/opt/adb/openshift/get_ip_address`
     echo
     echo "Successfully started and provisioned VM with #{VM_CPU} cores and #{VM_MEMORY} MB of memory."
     echo "To modify the number of cores and/or available memory set the environment variables"
@@ -85,9 +85,11 @@ Vagrant.configure(2) do |config|
     echo
     echo "You can now access the OpenShift console on: https://${OSIP}:8443/console"
     echo
+    echo "To download and install OC binary, run:"
+    echo "vagrant service-manager install-cli openshift"
+    echo
     echo "To use OpenShift CLI, run:"
-    echo "$ vagrant ssh"
-    echo "$ oc login"
+    echo "$ oc login ${OS_IP}:8443"
     echo
     echo "Configured users are (<username>/<password>):"
     echo "openshift-dev/devel"


components/rhel/rhel-ose/Vagrantfile
```diff
@@ -87,8 +87,8 @@ Vagrant.configure(2) do |config|
     echo "You can now access the OpenShift console on: https://${OSIP}:8443/console"
     echo
     echo "To use OpenShift CLI, run:"
-    echo "$ vagrant ssh"
-    echo "$ oc login"
+    echo "vagrant service-manager install-cli openshift"
+    echo "$ oc login ${OSIP}:8443"
```
brgnepal: I think we should rename `OSIP` to `OS_IP` which reads better and at least give user hint that is it IP of OS (openshift)

components/centos/centos-openshift-setup/Vagrantfile
```diff
@@ -115,8 +115,8 @@ Vagrant.configure(2) do |config|
     echo "Configured cluster admin user: admin, Password: admin"
     echo
     echo "To use OpenShift CLI, run:"
-    echo "$ vagrant ssh"
-    echo "$ oc login"
+    echo "vagrant service-manager install-cli openshift"
```
brgnepal, line 118: This may not be required if the user already has `oc` binary.

Probably, changing the wording with conditional phrase might make sense.

components/rhel/rhel-ose/Vagrantfile
LalatenduMohanty: Done

components/centos/centos-openshift-setup/Vagrantfile
LalatenduMohanty: Changed it to `To download OC binary : vagrant service-manager install-cli openshift`

brgnepal: How about this?

```
echo "If you don't have OpenShift CLI, oc, install it as:"
```

So that user will take the action accordingly.

LalatenduMohanty: @budhrg I am not sure if `If you don't have OpenShift CLI, oc, install it as:` is better than `To download and install OC binary, run:`.  I think we should ask others.  @Preeticp @thatdocslady comments?

components/centos/centos-openshift-setup/Vagrantfile
```diff
@@ -114,9 +114,11 @@ Vagrant.configure(2) do |config|
     echo
     echo "Configured cluster admin user: admin, Password: admin"
     echo
+    echo "To download and install OC binary, run:"
+    echo "vagrant service-manager install-cli openshift"
+    echo
     echo "To use OpenShift CLI, run:"
-    echo "$ vagrant ssh"
-    echo "$ oc login"
+    echo "$ oc login ${OPENSHIFT_HOSTNAME}:8443"
```
brgnepal, line 121: :+1:  Ya was missing it before.

components/rhel/rhel-ose/Vagrantfile
```diff
@@ -86,8 +86,10 @@ Vagrant.configure(2) do |config|
     echo "You can now access the OpenShift console on: https://${OSIP}:8443/console"
     echo
     echo "To use OpenShift CLI, run:"
-    echo "$ vagrant ssh"
-    echo "$ oc login"
+    echo "$ oc login ${OS_IP}:8443"
```
brgnepal: :+1:  Ya was missing it before.

components/rhel/rhel-ose/Vagrantfile
```diff
@@ -86,8 +86,10 @@ Vagrant.configure(2) do |config|
...
+    echo
+    echo "To download and install OC binary, run:"
+    echo "vagrant service-manager install-cli openshift"
```
LalatenduMohanty, line 89: Is `To download and install OC binary, run:` is ok or should we make it as `If you don't have OpenShift CLI, OC, install it as` 

@Preeticp @thatdocslady comments?

components/centos/centos-openshift-setup/Vagrantfile
brgnepal: Yes. Why I proposed that because, when instructions are visible to the command line, the users will do that as first thing. He might already have OpenShift binary which is already set and can be used. Then, he will decide whether to install it or not.

Probably better wordings will help here.

components/rhel/rhel-ose/Vagrantfile
Preeticp: If I have understood this correctly, how about putting the "To download and install OC binary, run... step before the "To use OpenShift CLI, run...step?
That way a new user would first download, install and then use, and a user who already has it will simply skip downloading and installing and move on to using it.

LalatenduMohanty: @Preeticp sounds good to me. @budhrg ^

brgnepal: :+1:

praveenkumar:
@LalatenduMohanty @budhrg what stopping this PR for merge? Looks like all reviews are done so we are in better position to merge it.

LalatenduMohanty:
@praveenkumar Will fix the PR and then we can merge it.

LalatenduMohanty:
@praveenkumar Will fix the PR and then we can merge it.

LalatenduMohanty:
@praveenkumar @budhrg it is ready to be merged.

praveenkumar:
LGTM :+1:

PR #541, opened 2016-09-12
Fix #540 ~/.vagrant.d/Vagrantfile does not affect vagrant-registration

### LINKED ISSUE #540: /.vagrant.d/Vagrantfile does not affect vagrant-registration
It seems that credentials stored in "~/.vagrant.d/Vagrantfile" does not apply to vagrant-registration plugin.

Code I am using in ~/.vagrant.d/Vagrantfile is (code1):

``` ruby
Vagrant.configure('2') do |config|
  config.registration.username = 'user'
  config.registration.password = 'login'
end
```

Code which is used for RHEL-OSE in CDK 2.2 (code2):

``` ruby
Vagrant.configure('2') do |config|
  ...  
  # vagrant-registration
  if ENV.has_key?('SUB_USERNAME') && ENV.has_key?('SUB_PASSWORD')
    config.registration.username = ENV['SUB_USERNAME']
  end
  ...
end
```

Despite the fact there are not any environment variables SUB_USERNAME or SUB_PASSWORD setup in my system, per system credentials in /.vagrant.d/ are ignored. I also tried to delete #vagrant-registration block above (code2) to be sure it is not overwritten by env variables, but problem still persists.
However when I put code from ~/.vagrant.d/Vagrantfile to RHEL-OSE Vagrantfile it works.

More CDK related info at: https://issues.jboss.org/browse/CDK-17?filter=-2

brgnepal:
Fix #540 
- IF there is no proxy provided through environment variables then
  don't invoke the "config.registration.*" calls to set them empty
  as it is changing the default vagrant behavior of getting those
  configurations from other Vagrantfile like ~/.vagrant.d/Vagrantfile
- Also, the current Vagrantfile assumes that proxy variables will be passed through environment variables. 
  However, one can change the Vagrantfile to put hard coded entries as below which is not at all recommended.

```
# check proxy_url, proxy_user, proxy_password

config.registration.proxy = PROXY = (ENV['PROXY'] || 'proxy_url' )
config.registration.proxyUser = PROXY_USER = (ENV['PROXY_USER'] || 'proxy_user')
config.registration.proxyPassword = PROXY_PASSWORD = (ENV['PROXY_PASSWORD'] || 'proxy_password')
```

**NOTE:** There will not be such way to put hard coded entries in Vagrantfile through this PR implementation.

### DIFF
--- a/components/rhel/misc/rhel-k8s-singlenode-setup/Vagrantfile
+++ b/components/rhel/misc/rhel-k8s-singlenode-setup/Vagrantfile
@@ -23,9 +23,9 @@ Vagrant.configure(2) do |config|
   end
 
   # Proxy Information from environment
-  config.registration.proxy = PROXY = (ENV['PROXY'] || '')
-  config.registration.proxyUser = PROXY_USER = (ENV['PROXY_USER'] || '')
-  config.registration.proxyPassword = PROXY_PASSWORD = (ENV['PROXY_PASSWORD'] || '')
+  ENV['PROXY'] ? (config.registration.proxy = PROXY = ENV['PROXY']) : PROXY = ''
+  ENV['PROXY_USER'] ? (config.registration.proxyUser = PROXY_USER = ENV['PROXY_USER']) : PROXY_USER = ''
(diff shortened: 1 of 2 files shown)


optak:
I don't understand the code entirely and didn't try it, but patch looks good to me.
Summary of use cases this should cover (both for general Vagrantfile, without changing local one):
- skip registration
- hardcoded username and password

LalatenduMohanty:
LGTM

PR #549, opened 2016-09-28
Fix #548 update OSE tag

### LINKED ISSUE #548: Update ose tag with 3.3 release
Currently we still have 3.2 release tag for OSE and v3.3.0.32 is now available at registry.access.redhat.com, so it need to be updated.

praveenkumar:
This patch will update OSE tag which already released.


LalatenduMohanty:
LGTM

PR #551, opened 2016-09-29
Fix #550 add .cdk file should be in cdk.zip for devsuite

### LINKED ISSUE #550: cdk.zip should include a .cdk file in same folder as Vagrantfile
Content of this file should be the following type:

openshift.auth.scheme=Basic
openshift.auth.username=openshift-dev
openshift.auth.password=devel

It would also be nice if we were able to include a version here, but as I understand it, the version of the CDK is based on the box that's installed, yes? Not the cdk.zip file or the Vagrantfile?

However, if you were able to add a versioning key here, if we could assume a user is downloading CDK as one, it would be very helpful for our usecase.

Our usecase in JBossTools is to automatically detect CDK installations. Issue is to be linked.

Source: https://issues.jboss.org/browse/CDK-15

praveenkumar:
Fix #550

### DIFF
--- a/components/rhel/rhel-ose/.cdk
+++ b/components/rhel/rhel-ose/.cdk
@@ -1,0 +1,4 @@
+openshift.auth.scheme=Basic
+openshift.auth.username=openshift-dev
+openshift.auth.password=devel
+cdk.box.version=2.3


LalatenduMohanty:
LGTM

components/rhel/rhel-ose/.cdk
```diff
@@ -0,0 +1,4 @@
+openshift.auth.scheme=Basic
+openshift.auth.username=openshift-dev
+openshift.auth.password=devel
+cdk.box.version=2.3
```
brgnepal, line 4: @praveenkumar Why the box version is 2.3 here? Does it represent the upcoming version?

PR #563, opened 2016-10-14
Cleaned up formatting and other post-ascii-doc conversion fixes

### LINKED ISSUE #560: Docs Cleanup post asciidoc conversion
This Issue lists cleanup tasks in the Installing and Using docs, resulting from conversion of docs to Asciidoc:
1. installing.adoc: This doc had numbered second level headings as they are actually Steps to be run in an order.  The last heading: Setup ADB behind proxy however has no numbering,  I am wondering if this should be added before the Vagrant up step or maybe a note. WDYT?
2. using.adoc: 
- Using ADB with host-based tools (Eclipse and CLIs); Line36 has a dashed line (RAW format)which makes the entire doc below it grey, though on rendering I see no problem.
- Step 7. If you are using Eclipse, you should follow these steps:
  Install the Docker Tooling plugin. ,.  Enable the three Docker Views (Docker Explorer, Docker Containers, and Docker Images) by choosing Windows→Show Views→Others.
  These are 2 steps and the dot before Enable is for auto numbering in adoc.
- Windows→Show Views→Others, Docker explorer, connection name are all rendered differently let us fix one style for these GUI Labels. Maybe a simple Bold text for all GUI Labels? (I think this is a legacy issue, and not a result of the conversion. Just noting it here so that we know that this issue exists.)
- Alignment of Note in using.adoc step 7 needs to be corrected.
- Line 198 sccli readme is in rst, but has been changed to adoc. So do we revert to rst as link is broken currently, until there is a decision on the Plugin Conversions #557 

We can add other tasks here for other docs as the reviews happen.

thatdocslady:
Fixes #560 

This PR includes various formatting, markup, structure, and presentation fixes that were needed after the asciidoc conversion.

### DIFF
--- a/docs/installing.adoc
+++ b/docs/installing.adoc
@@ -9,7 +9,7 @@ toc::[]
 '''
 
 [[install-a-virtualization-provider]]
-== Install a virtualization provider
+== Step 1. Install a virtualization provider
 
 Two virtualization providers have been tested with ADB; VirtualBox and
 Libvirt.
@@ -92,7 +92,7 @@ automatically installed together with the CentOS software collections
 build.
 
 [[install-vagrant]]
-== Install Vagrant
+== Step 2. Install Vagrant
 
 * Microsoft Windows
 1.  Follow the directions at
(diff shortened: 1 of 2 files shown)


thatdocslady:
@LalatenduMohanty @praveenkumar can you pleaser review and merge if ok?

LalatenduMohanty:
LGTM

PR #565, opened 2016-10-14
Fix #564; added note wrt name of CDK box must be same as that in Vagrantfile.

### LINKED ISSUE #564: README.rst should specifiy that Vagrantfile expects the box file to be called cdkv2
In the README.rst, which is included in the cdk.zip, you're told to download the box image next to the Vagrantfile (into the same folder) and then to launch it via vagrant up. 
The README.rst is missing to specifiy though, that the image should be called cdkv2. One finds out with the error that's displayed in the terminal.
Symlinking my cdkv2 -> rhel-7-cdk-vagrant-scratch-7.2.07102016-1.x86_64.vagrant-libvirt.box fixed the issue for me.

Preeticp:
This PR fixes #564.
1. Adds note about the CDK name to correspond with the Vagrantfile used to initialize the box.
2. Reversed the order of the first 2 steps to imply that: We need to create a folder first and then download CDK and the Vagrantfile in the same folder.
3. Provided link to detailed CDK Installation guide.
4. Added TOC
This has been done for both Readmes.

### DIFF
--- a/components/rhel/misc/rhel-k8s-singlenode-setup/README.adoc
+++ b/components/rhel/misc/rhel-k8s-singlenode-setup/README.adoc
@@ -1,5 +1,6 @@
 = Vagrantfile: Single-Node Kubernetes
 :toc:
+:toc-placement!:
 
 This Vagrantfile is the suggested configuration for using the CDK with
 Kubernetes. This file sets up private networking that will be used to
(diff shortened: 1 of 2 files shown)


praveenkumar:
LGTM +1

PR #580, opened 2016-11-08
Remove rst format doc now adoc is available for same

praveenkumar:
We have adoc variant available for same so removing this.


Preeticp:
LGTM

PR #586, opened 2016-11-15
Fix #570 Consistent format for setting Vagrantfile configurations

### LINKED ISSUE #570: Consistent format for setting Vagrantfile configurations
Currently, we are using two formats in Vagrantfile for setting configurations
- Using `if` block as

```
if ENV.has_key?('SUB_USERNAME') && ENV.has_key?('SUB_PASSWORD')
    config.registration.username = ENV['SUB_USERNAME']
end
```
- One liner approach as

```
ENV['PROXY'] ? (config.registration.proxy = PROXY = ENV['PROXY']) : PROXY = ''
```

First one gives more readability while second option allow to achieve minimal Vagrantfile goal.

Any thoughts which should be used? 

**NOTE**:
`ENV['PROXY']` should be `ENV.key?('PROXY')` (Should be fix as part of this issue fix)

brgnepal:
Fix #570 

I tried to remove empty assignment of PROXY* constant but get following error on `vagrant destroy` which seems to be valid.
```
/home/budhram/redhat/vagrant-service-manager/Vagrantfile:87:in `block in <top (required)>': uninitialized constant PROXY (NameError)
..............
```

### DIFF
--- a/components/rhel/misc/rhel-k8s-singlenode-setup/Vagrantfile
+++ b/components/rhel/misc/rhel-k8s-singlenode-setup/Vagrantfile
@@ -18,18 +18,11 @@ end
 
 # Vagrantfile for single node k8s setup
 Vagrant.configure(2) do |config|
-  # Use environment variable BOX or default 'cdkv2'
-  config.vm.box = (ENV.key?('BOX') ? (ENV['BOX'].empty? ? BOX_NAME : ENV['BOX']) : BOX_NAME)
-
-  if ENV.has_key?('SUB_USERNAME') && ENV.has_key?('SUB_PASSWORD')
-    config.registration.username = ENV['SUB_USERNAME']
-    config.registration.password = ENV['SUB_PASSWORD']
-  end
-
-  # Proxy Information from environment
-  ENV['PROXY'] ? (config.registration.proxy = PROXY = ENV['PROXY']) : PROXY = ''
-  ENV['PROXY_USER'] ? (config.registration.proxyUser = PROXY_USER = ENV['PROXY_USER']) : PROXY_USER = ''
-  ENV['PROXY_PASSWORD'] ? (config.registration.proxyPassword = PROXY_PASSWORD = ENV['PROXY_PASSWORD']) : PROXY_PASSWORD = ''
+  config.vm.box = if ENV.key?('BOX')
+                  ENV['BOX'].empty? ? BOX_NAME : ENV['BOX']
+                else
+                  BOX_NAME
+                end
 
   config.vm.provider "libvirt" do |libvirt, override|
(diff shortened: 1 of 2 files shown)


components/rhel/misc/rhel-k8s-singlenode-setup/Vagrantfile
```diff
@@ -41,8 +37,23 @@ Vagrant.configure(2) do |config|
...
+
+  # Proxy Information from environment
+  if ENV.key?('PROXY')
+    config.registration.proxy = PROXY = ENV['PROXY']
+    config.registration.proxyUser = PROXY_USER = ENV['PROXY_USER']
+    config.registration.proxyPassword = PROXY_PASSWORD = ENV['PROXY_PASSWORD']
+  else
+    PROXY = PROXY_USER = PROXY_PASSWORD = ''
+  end
+
```
praveenkumar, line 62: What if we only have proxy server without authentication so how this will work out. does it set `PROXY_USERNAME` and `PROXY_PASSWORD` as `''`?

components/rhel/misc/rhel-k8s-singlenode-setup/Vagrantfile
```diff
@@ -41,8 +37,23 @@ Vagrant.configure(2) do |config|
...
+  if ENV.has_key?('SUB_USERNAME') && ENV.has_key?('SUB_PASSWORD')
+    config.registration.username = ENV['SUB_USERNAME']
+    config.registration.password = ENV['SUB_PASSWORD']
+  end
+
+  # Proxy Information from environment
+  if ENV.key?('PROXY')
+    config.registration.proxy = PROXY = ENV['PROXY']
+    config.registration.proxyUser = PROXY_USER = ENV['PROXY_USER']
+    config.registration.proxyPassword = PROXY_PASSWORD = ENV['PROXY_PASSWORD']
```
LalatenduMohanty: This assumption is not correct. You can have a proxy environment which do not have user name and password.

LalatenduMohanty (changes requested):

components/rhel/misc/rhel-k8s-singlenode-setup/Vagrantfile
brgnepal: Yes @praveenkumar , but what is expected in that case? 
Empty or null ?

praveenkumar: @budhrg it suppose to be empty because otherwise it is undefined. Also make sure you did testing for this change.

LalatenduMohanty (approved):

LalatenduMohanty:
LGTM

PR #589, opened 2016-11-17
Fix #588 No section for hyperv in OSE Vagrantfile

### LINKED ISSUE #588: No section for hyperv in OSE Vagrantfile
We don't have `hyperv` provider section in our Vagrantfile which not able to create VM with our desired RAM + CPU and our openshift setup fails.

praveenkumar:
This patch have fix for hyperv.


LalatenduMohanty:
LGTM

PR #594, opened 2016-11-21
Fixed section IDs and relative links to other docs

### LINKED ISSUE #553: Cleanup tasks to prepare for docs publication
Following the asciidoc conversion, there are some tasks that still need to be performed on the overall doc set in order to get it ready for publication:
- Update and formalize permalink IDs (for each h2, h3). Currently they're auto-generated in the asciidoc files, but because they will ultimately end up in the URL breadcrumbs it's probably a good idea to review them and make sure they're concise and accurate (it'll be harder to change later and if we can avoid redirects that's a good thing (tm) ).
- Cross-document (and some cross-repo) links currently refer to the adoc files. We need to replace these with the published URLs (this should probably wait until we do a test publication and have the full URL), otherwise users will end up in the GH files instead of the published content.

thatdocslady:
Fixes #553 

This PR addresses the following issues:

- Fix section/topic IDs after asciidoc conversion (many of them were long and clumsy)
- Fix links to other docs in the repo to be relative (instead of hard-coded) to allow them to parse into HTML links when we publish docs later
- Fixed links that were broken due to line breaks 
- Minimize links to other repos (only a handful of them are currently used and only when absolutely necessary)

Reviews welcome!

### DIFF
--- a/docs/installing.adoc
+++ b/docs/installing.adoc
@@ -8,7 +8,7 @@ This document describes how to install ADB and required dependencies.
 toc::[]
 '''
 
-[[install-a-virtualization-provider]]
+[[install-virt-provider]]
 == Step 1. Install a virtualization provider
 
 Two virtualization providers have been tested with ADB; VirtualBox and
@@ -34,7 +34,7 @@ The choice as to which to use should be driven by your preferences and
 environmental concerns and is outside of the scope of this document.
 Both will work equally well in their default configuration. You may wish
 to read the section on file synchronization in the
-https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/using.adoc[Using ADB documentation] when making this decision.
+link:using.adoc[Using ADB documentation] when making this decision.
 
 * VirtualBox installation instructions are available
 https://www.virtualbox.org/manual/ch02.html#startingvboxonlinux[online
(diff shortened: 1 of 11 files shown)


docs/installing.adoc
```diff
@@ -252,6 +252,30 @@ $ curl -sL https://raw.githubusercontent.com/projectatomic/adb-atomic-developer-
 Refer:
 link:../components/centos/centos-mesos-marathon-singlenode-setup/README.adoc[README]
 
+<<<<<<< a7f6568029db13e9f431b881a8ba11942dd28eb4
```
LalatenduMohanty: @thatdocslady Please fix this. I think this is a byproduct of rebase.

docs/installing.adoc
```diff
@@ -285,7 +309,7 @@ export BOX=<box-name>
 ----
 ====
 
-[[set-up-adb-behind-an-http-proxy]]
+[[setup-http-proxy]]
 == Step 6. (Optional) Set up ADB behind an HTTP proxy
```
LalatenduMohanty, line 289: `setup-http-proxy` looks incorrect. The section represents `how to setup ADB behind a proxy setup`

LalatenduMohanty (changes requested):

thatdocslady:
@LalatenduMohanty fixed the sections per your feedback, thanks! @hferentschik @Preeticp can you also please review? Cheers :)

LalatenduMohanty (approved):

LalatenduMohanty:
@Preeticp waiting for your review. Thanks.

Preeticp:
There is just this one substep in step 5, in the Installation guide, being discussed, apart from that, the rest LGTM.

thatdocslady:
@Preeticp fixed step 5 in the install doc per our discussion.

Preeticp:
LGTM @thatdocslady, @LalatenduMohanty

PR #607, opened 2017-01-23
Fix #606 Use PROXY* as variable instead of constant

### LINKED ISSUE #606: Use PROXY* as variable instead of constant
`PROXY*` entities are declared as constant in Vagrantfile but used as variable like [here](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/components/rhel/rhel-ose/Vagrantfile#L69-L71).

This will have side effect when exporting these entities via terminal as constants in Ruby as global.

```
/home/budhram/redhat/vagrant-service-manager/Vagrantfile:69: warning: already initialized constant PROXY
/home/budhram/redhat/vagrant-service-manager/Vagrantfile:69: warning: previous definition of PROXY was here
/home/budhram/redhat/vagrant-service-manager/Vagrantfile:94:in `block in <top (required)>': uninitialized constant PROXY_USER (NameError)
	from /home/budhram/.rvm/gems/ruby-2.1.2/bundler/gems/vagrant-bedaeee190b2/lib/vagrant/config/v2/loader.rb:37:in `call'
	from /home/budhram/.rvm/gems/ruby-2.1.2/bundler/gems/vagrant-bedaeee190b2/lib/vagrant/config/v2/loader.rb:37:in `load'
........
```

brgnepal:
Fix #606

### DIFF
--- a/components/centos/centos-docker-base-setup/Vagrantfile
+++ b/components/centos/centos-docker-base-setup/Vagrantfile
@@ -67,5 +67,4 @@ Vagrant.configure(2) do |config|
   config.vm.provision "shell", inline: <<-SHELL
     sudo setsebool -P virt_sandbox_use_fusefs 1
   SHELL
-  # config.servicemanager.services = 'docker'
 end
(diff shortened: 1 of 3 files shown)


praveenkumar:
LGTM :+1:

PR #609, opened 2017-01-24
Fix #608 Update OCP tag and kickstart file

### LINKED ISSUE #608: Update OCP version and kickstart
We need to update OCP version and kickstart file for this release.

praveenkumar:
This patch is to update kickstart file and OCP tags.


brgnepal:
LGTM :+1:

